PR #6, opened 2023-06-01
Experiment: produce PDF with Puppeteer

ralfhandl:
This is currently only a sketch of how the final solution could look like.

Still missing:
- [x] parameterize with filename and variables from meta.yaml
- [x] integrate into Github workflow
- [x] only generate PDF if HTML or Markdown have changed
- [ ] extract inline styles to `odata.css`?


HeikoTheissen:
> integrate into Github workflow

No need to run this with every push, unless we want to publish a PDF on github.io.

ralfhandl:
Always having the PDF in sync with the HTML would allow spotting PDF production problems early.

docs/odata-data-aggregation-ext/odata-data-aggregation-ext.pdf
HeikoTheissen, line 1: Bigger margin above the footer, please.

ralfhandl: Please check if the additional 20% are sufficient

HeikoTheissen: Have you regenerated the PDF after increasing the margin? I don't see an improvement yet.

HeikoTheissen (approved):

PR #14, opened 2023-06-19
PDF: explicitly set return code 0

ralfhandl:


HeikoTheissen (commented):
The successful workflow run is not reported back to the PR.

HeikoTheissen (approved):
Let's see whether it helps.

PR #21, opened 2023-06-30
Chore: bump c8 and puppeteer to latest

ralfhandl:


HeikoTheissen (approved):
Appoved after bumping semver again

PR #27, opened 2023-07-06
Add byte-order mark to docs/*/*.md file

HeikoTheissen:
By adding a byte-order mark to the generated Markdown files, they are displayed as UTF-8 when loaded from github.io.

A browser in a Windows environment displays step 8 in section 3.1.3 of https://oasis-tcs.github.io/odata-specs/odata-data-aggregation-ext/odata-data-aggregation-ext.md (the authoritative document!) wrongly as:
```
8. Let $B=Î³(v,p_2)$.
```

With the byte-order mark, it correctly displays:
```
8. Let $B=γ(v,p_2)$.
```

### DIFF
--- a/docs/odata-data-aggregation-ext/odata-data-aggregation-ext.md
+++ b/docs/odata-data-aggregation-ext/odata-data-aggregation-ext.md
@@ -1,4 +1,4 @@
-
+﻿
 ![OASIS Logo](https://docs.oasis-open.org/templates/OASISLogo-v3.0.png)
 
 -------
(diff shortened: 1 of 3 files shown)


ralfhandl:
How does that affect for example Mac users?

HeikoTheissen:
> How does that affect for example Mac users?

If you have a Mac, visit https://heikotheissen.github.io/markdowntest/charset.md. You should see
```
Heiko Theißen
```
although the `Content-Type` header does not specify a `charset`.

ralfhandl (approved):

PR #30, opened 2023-07-14
Fixes requested by TC Admin

ralfhandl:
- [x] Date should be set to 05 July 2023 (In the subtitle, "Citation format", and in the footers in PDF)
- [x] in the PDF, the file name in the footer should include the "v4.0" - thus "odata-data-aggregation-ext-v4.0-csd04"
- [x] The HTML, MD, and PDF file names (i.e., the files inside the ZIP) should include the Version and stage: 
  - odata-data-aggregation-ext-v4.0-csd04.html (or .md or pdf)


GeraldKrause (approved):
Looks good to me. First point of the review feedback obviously did not require a fix.

PR #48, opened 2023-07-26
CSDL JSON & XML: clean 4.01 OS documents for copy-editing

ralfhandl:
- [x] download 4.01 HTML
- [x] add scripts to package.json
- [x] check CSDL JSON and improve clean.mjs
- [x] check CSDL XML and improve clean.mjs


HeikoTheissen (dismissed):

HeikoTheissen:
So far we distinguish between 4.0 and 4.01 payloads. Must we start to consider 4.02 payloads as well if we publish ODATA-CSDL v4.02?

ralfhandl:
> So far we distinguish between 4.0 and 4.01 payloads. Must we start to consider 4.02 payloads as well if we publish ODATA-CSDL v4.02?

We should strive for 4.0 vs. "4.01 or later" and avoid different behaviour between 4.01 and 4.02.

HeikoTheissen (approved):

ralfhandl:
@HeikoTheissen do you want to merge?

PR #75, opened 2023-08-01
Example sale 8 is for product P3, not P1

HeikoTheissen:
While transferring the example data section from Word to Markdown, a typo was introduced: Example sale 8 must be for product P3, not for P1. The error affects the Sales table in section 2.2 and the non-cross table in section 2.3. The rest of the document, including all examples, are correctly based on sale 8 with product P3.

See also https://lists.oasis-open.org/archives/odata-comment/202308/msg00000.html


GeraldKrause (approved):
LGTM

GeraldKrause (dismissed):
LGTM

PR #84, opened 2023-08-03
Tools/build pdfs only when needed

### LINKED ISSUE #56: Node CI: only build PDFs whose html/css sources have changed
After building and adding html and md:
- Loop over folders in ˋdocsˋ
  - ˋdiff-indexˋ only this folder & build the PDF in this folder if necessary
- Add & commit all changes

- [ ] parameterize pdf script
- [ ] adapt workflow

ralfhandl:
Fixes #56

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -24,11 +24,19 @@ jobs:
       - name: Build docs
         run: |
           npm ci
-          npm run build --if-present
+          npm run build
           git config user.name ${GITHUB_ACTOR}
           git config user.email ${PUSHER_EMAIL}
           git add docs/*/*.html docs/*/*.md docs/*/styles/*
-          git diff-index --quiet HEAD docs/*/*.html docs/*/*.md docs/*/styles/* || (npm run pdf --if-present && git add docs/*/*.pdf && git commit -m "auto-refreshed")
+          for dir in docs/*/ # list directories in the form "docs/dirname/"
+          do
+              doc=${dir%/}   # remove the trailing "/"
+              doc=${doc##*/} # retain everything after the final "/"
+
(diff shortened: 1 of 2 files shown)


HeikoTheissen (dismissed):

.github/workflows/nodejs.yml
```diff
@@ -24,11 +24,17 @@ jobs:
...
-          git diff-index --quiet HEAD docs/*/*.html docs/*/*.md docs/*/styles/* || (npm run pdf --if-present && git add docs/*/*.pdf && git commit -m "auto-refreshed")
-          test `git branch --show-current` && git push
+          for dir in docs/*/ # list directories in the form "docs/dirname/"
+          do
+              doc=${dir%/}   # remove the trailing "/"
+              doc=${doc##*/} # retain everything after the final "/"
+              git diff-index --quiet HEAD $dir/*.html $dir/styles/* || npm run pdf $doc
+          done
+          git add docs/*/*.pdf
+          git diff-index --quiet HEAD || ( git commit -m "auto-refreshed" && git push )
```
HeikoTheissen, line 38: A non-selective `git diff-index` after a selective `git add` always returns 1, see https://github.com/SAP/odata-vocabularies/pull/271.

An unconditional `git push` leads to an error when not on a branch (for example, when this check runs after pushing a tag).

HeikoTheissen (changes requested):

.github/workflows/nodejs.yml
ralfhandl: `git diff-index` is now selective

ralfhandl:
> The improvement to `build-pdf.js` is worth merging, however.

#85, please review & merge

HeikoTheissen (approved):

PR #97, opened 2023-08-15
Automated HTML abstracts

### LINKED ISSUE #31: HTML: add meta tag with name="description" and text of the abstract
For Data Aggregation this would be
~~~html
  <meta name="description" content="This specification adds basic grouping and aggregation functionality (e.g. sum, min, and max) to the Open Data Protocol (OData).">
~~~
right before
~~~html
  <meta charset="utf-8" />
~~~

This would reduce the manual work for OASIS TC Admin.

HeikoTheissen:
Fixes #31


odata-json-format/meta.yaml
```diff
@@ -6,3 +6,9 @@ previousStage: n/a
 pubdate: 14 July 2023
 pubdateISO: '2023-07-14'
 copyright: © OASIS Open 2023
+description:
+  "The Open Data Protocol (OData) for representing and interacting with structured
```
ralfhandl, line 10: Why do we need the double-quotes here?

ralfhandl (commented):
Fine, just one question regarding meta.yaml in odata-json-format

odata-json-format/meta.yaml
HeikoTheissen: Otherwise the row
```
protocol is in OData Version 4.02 Part 1: Protocol.
```
is interpreted as YAML key-value pair.

ralfhandl: [Block Scalars](https://yaml-multiline.info/) seem to do the trick

ralfhandl (approved):

PR #134, opened 2023-08-18
ODATA-1432

ralfhandl:


HeikoTheissen (changes requested):

HeikoTheissen (dismissed):

odata-csdl/7 Structural Property.md
```diff
@@ -155,10 +155,10 @@ value
 
 ## ##subsec Type Facets
 
-Facets modify or constrain the acceptable values of a property.
+Facets modify or constrain the acceptable values for a model element of a given type.
 
-For single-valued properties the facets apply to the value of the
-property. For collection-valued properties the facets apply to the items
+For single-valued model elements the facets apply to the value of the
```
mikepizzo, line 160: As this is within the definition of a property, it seems we can be clearer by using "property" here.

odata-csdl/7 Structural Property.md
```diff
@@ -155,10 +155,10 @@ value
 
 ## ##subsec Type Facets
 
-Facets modify or constrain the acceptable values of a property.
+Facets modify or constrain the acceptable values for a model element of a given type.
```
mikepizzo, line 158:
```suggestion
Facets applied to a property modify or constrain the acceptable values for the property.
```

odata-csdl/1 Introduction.md
```diff
@@ -488,6 +488,306 @@ representation of primitive type values in URLs and
...
+
+Absence of `$Scale` means `variable`.
+:::
+
+::: {.varjson .example}
+Example ##ex: [`Precision`](#Precision)`=3` and `Scale=2`.  
+Allowed values: 1.23, 0.23, 3.14 and 0.7, not allowed values: 123, 12.3
+```json
+"Amount32": {
+  "$Nullable": true,
```
HeikoTheissen, line 648: Nullable is introduced only in section 7.2, can we avoid it in these early examples?

ralfhandl: Well, if we want to have isomorphic examples for JSON and XML, then one of them has to talk about Nullable because they have opposite defaults.

Where do we want to have Nullable: in JSON or in XML?

lib/number.js
ralfhandl, line 1: Should we pull this out into a separate tooling PR?

package.json
```diff
@@ -22,5 +22,8 @@
   "devDependencies": {
     "connect-livereload": "^0.6.1",
     "livereload": "^0.9.3"
-  }
+  },
+	"prettier": {
+		"trailingComma": "none"
+	}
```
ralfhandl, line 28: I'd rather stick with the default settings of the latest Prettier.

odata-csdl/1 Introduction.md
HeikoTheissen: Leave the examples as they are but add a comment: "The `$Nullable` member is explained in section ...".

HeikoTheissen (approved):

HeikoTheissen (dismissed):

odata-csdl/1 Introduction.md
```diff
@@ -494,6 +494,305 @@ representation of primitive type values in URLs and
 [OData-JSON](#ODataJSON) for the representation in requests and
 responses.
 
+## ##subsec Type Facets
+
+Facets modify or constrain the acceptable values for a model element of a given primitive type,
```
mikepizzo, line 499: I initially read this as the model element belonging to a given primitive type, not that the model element was a given primitive type.

Perhaps turn this around; something like:
```suggestion
Facets modify or constrain the acceptable values of primitive typed model elements,
```

mikepizzo (changes requested):
Looks good overall -- one small comment with the wording in the introduction.

mikepizzo (approved):

HeikoTheissen (dismissed):

mikepizzo (approved):

PR #138, opened 2023-08-18
ODATA-1622: Avoid "MAY ... only if"

HeikoTheissen:
Combining the RFC2119 term "MAY" with a negation or restrictions runs counter to what MAY wants to express.
> API designers MAY do A only if B

sounds as if
* the API designer may choose not to do A if B (which is not always implied)
* there is no rule for the client if not B.

Instead, write
> API designers MUST NOT do A if (not B)

https://issues.oasis-open.org/browse/ODATA-1622


ralfhandl (approved):

ralfhandl (approved):

sthagen (approved):
LGTM (resolving ODATA-1622, I guess)

ralfhandl (approved):

PR #142, opened 2023-08-18
ODATA-1482

ralfhandl:


odata-protocol/8 Header Fields.md
```diff
@@ -1035,7 +1035,9 @@ include a response body describing the functionality not implemented.
 
 ## ##subsec Error Response Body
 
-The representation of an error response body is format-specific. It
+The representation of an error response body is format-specific.
+An OData-specific error response (which can be recognized by the presence
+of the [`OData-Version`](#HeaderODataVersion) header)
```
HeikoTheissen, line 1040:
```suggestion
An error response body can be the result of a failure of OData processing or of the underlying protocol.
An OData-specific error response (which can be recognized by the presence
of the [`OData-Version`](#HeaderODataVersion) header) is format-specific and
```

ralfhandl: "underlying infrastructure" or "other software layers"? Need not be protocol-related.

HeikoTheissen (approved):

PR #145, opened 2023-08-18
ODATA-1495

ralfhandl:
https://issues.oasis-open.org/browse/ODATA-1495


HeikoTheissen (approved):

GeraldKrause (commented):
The change now refers to the format-specific representation also when creating an entity. This was not discussed in https://issues.oasis-open.org/browse/ODATA-1495. 
Is this generally true for creating entities, also without an OData-Version header with a value of 4.01?

ralfhandl:
> The change now refers to the format-specific representation also when creating an entity. This was not discussed in https://issues.oasis-open.org/browse/ODATA-1495. 

I took the liberty of globally searching for "base64url-encoded" and replacing it in both places with "format-specific", which I think is correct.

> Is this generally true for creating entities, also without an OData-Version header with a value of 4.01?

Good question, could you please create a new issue for it?

GeraldKrause:
Alright, I will create a new Jira issue for this question. Let's put this PR on hold until we have clarified this point, because the current change assumes that the rule applies for creating entities also without the OData-Version header.

HeikoTheissen (approved):

HeikoTheissen (approved):

PR #147, opened 2023-08-18
ODATA-1510

ralfhandl:


odata-url-conventions/5 Query Options.md
```diff
@@ -1685,12 +1684,19 @@ http://host/service/Employees?$filter=LastName eq $root/Employees('A1245')/LastN
...
+`$root` to each customer
 ```
 http://host/service/ProductsOrderedBy(Customers=@c)?@c=[$root/Customers('ALFKI'),$root/Customers('BLAUS')]
 ```
 :::
 
+::: example
+Example ##ex: function call returning the average rating of a given employee by their peers (employees in the same department)
+```
+http://host/service/Employees('A1245')/self.AvgRating(RatedBy=@peers)?@peers=$root/Employees/$filter(Department eq $it/Department)
```
ralfhandl, line 1696: What would `$it/Department` be? The resource path includes the function call whose result depends on the value of `@peers`.

Better choose a more direct example

Example copied from https://www.oasis-open.org/apps/org/workgroup/odata/download.php/69589/odata-v4.01-errata01-os-part2-url-conventions.docx.

HeikoTheissen: > What would the value of `$it/Department` be? The resource path includes the function call whose result depends on the value of `@peers`

The resource path to which `$it` refers is `$root/Employees` in this case, not `http://host/service/Employees('A1245')/self.AvgRating(RatedBy=@peers)`.

ralfhandl: According to https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_URLComponents the resource path is everything after the service root up to (and excluding) the first question mark or hash, or up to the end of the URL, that is `Employees('A1245')/self.AvgRating(RatedBy=@peers)`

ralfhandl: We could make it more complicated:
~~~
http://host/service/Employees('A1245')/self.AvgRating(RatedBy=@peers)
  ?@peers=$root/Employees/$filter(Department eq @dept)
  &@dept=$root/Employees('A1245')/Department
~~~

HeikoTheissen: > Ok, first nit: we don't specify that `$root` introduces a new context for `$it`.
> 
> Second nit: if we would specify that, then `Department` would be the same as `$it/Department` because the argument of `/$filter(...)` is evaluated in the context of the prepended path `$root/Employees`.

You are right, I withdraw my proposal. Take the more complicated example, or
```
http://host/service/Employees('A1245')/self.AvgRating(RatedBy=@peers)
  ?@peers=$root/Employees/$filter(Department eq 'D1')
```

HeikoTheissen, line 1696:
```suggestion
Example ##ex: function call returning the average rating of a given employee by their peers (employees in department D1)
```
```suggestion
http://host/service/Employees('A1245')/self.AvgRating(RatedBy=@peers)
?@peers=$root/Employees/$filter(Department eq 'D1')
```

HeikoTheissen (changes requested):

odata-url-conventions/5 Query Options.md
```diff
@@ -1685,12 +1684,19 @@ http://host/service/Employees?$filter=LastName eq $root/Employees('A1245')/LastN
...
 customers is passed as a JSON array containing the resource paths from
-`$root` to each customer.
+`$root` to each customer
 ```
 http://host/service/ProductsOrderedBy(Customers=@c)?@c=[$root/Customers('ALFKI'),$root/Customers('BLAUS')]
 ```
 :::
 
+::: example
+Example ##ex: function call returning the average rating of a given employee by their peers (employees in the same department)
```
HeikoTheissen, line 1694:
```suggestion
Example ##ex: function call returning the average rating of a given employee by their peers (employees in department D1)
```
The second change seems to have overridden the first, here it is again.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #151, opened 2023-08-18
ODATA-1525

ralfhandl:


odata-protocol/10 Context URL.md
```diff
@@ -310,6 +310,9 @@ parentheses is omitted from the select-list if it does not contain a
 nested `$select `or` $expand`, but MUST still be present, without a
 suffix, if it is explicitly selected.
 
+There is no shortcut for representing the list of all navigation properties;
+`$expand=*` is treated as if the corresponding navigation properties were explicitly expanded.
```
HeikoTheissen, line 314:
```suggestion
The context URL has no shortcut for representing the list of all navigation properties;
`$expand=*` is treated as if all navigation properties were explicitly expanded.
```
Is that what you mean?

HeikoTheissen (changes requested):

odata-protocol/10 Context URL.md
ralfhandl: Thanks, that is much clearer

HeikoTheissen (dismissed):

HeikoTheissen (approved):

PR #152, opened 2023-08-18
ODATA-1539

ralfhandl:


HeikoTheissen (changes requested):
ODATA-1539 was meant for ERRATA01, but now we are preparing V4.02 and don't need it, we can use ODATA-1540.

HeikoTheissen (dismissed):

HeikoTheissen (approved):

PR #156, opened 2023-08-21
Aggregation on cross-join

HeikoTheissen:
This pull request corrects example 48 in [OData-Data-Agg-v4.0]:
CS02|CSD04
---|---
`aggregate(Sales(Amount with sum as Total))`|`aggregate(Sales/Amount with sum as Total)`


GeraldKrause (approved):
LGTM

PR #159, opened 2023-08-23
ODATA-1446

ralfhandl:


HeikoTheissen (commented):
This resolution differs from the proposal in https://issues.oasis-open.org/browse/ODATA-1446.

PR #164, opened 2023-08-30
No dashed line as first line of markdown file

ralfhandl:
VS Code's markdown editor misinterprets a line of dashes at the beginning of a markdown file as the beginning of YAML frontmatter, which results in an incorrect or empty outline view.

This has no effect on the generated docs/*/*.html files.


HeikoTheissen (approved):
You have even inserted two missing hrules.

PR #170, opened 2023-09-08
ODATA-1405

### LINKED ISSUE #320: /$query shall allow Content-Type: application/x-www-form-urlencoded
An HTML form can be used to have a browser _navigate_ to the response of an OData request, especially if the response is a stream. If that OData request makes use of the [`/$query`](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody) segment, it must have `Content-Type: text/plain`, which forces an HTML form like
```xml
<form method="post" action="~/$query" enctype="text/plain">
 <input name="$filter" value="a%20eq%201&$select=b"/>
</form>
```
Much more natural would be if `Content-Type: application/x-www-form-urlencoded` were also allowed:
```xml
<form method="post" action="~/$query"
  enctype="application/x-www-form-urlencoded">
 <input name="$filter" value="a eq 1"/>
 <input name="$select" value="b"/>
</form>
```
This leads to the same request body as above:
```
$filter=a%20eq%201&$select=b
```
but without forcing the client to percent-encode spaces and insert the ampersand.

### Proposal

The request body MUST use the content-type `text/plain` _or `application/x-www-form-urlencoded`_.

https://github.com/oasis-tcs/odata-specs/pull/170

Imported from [ODATA-1405](https://issues.oasis-open.org/browse/ODATA-1405)

HeikoTheissen:
Fixes #320


odata-url-conventions/4 Resource Path.md
```diff
@@ -959,4 +959,34 @@ Content-Type: text/plain
...
+```
+:::
+
+::: example
+Example ##ex: passing multiple system query options in the request body
+```
+POST http://host/service/People/$query
+Content-Type: application/x-www-form-urlencoded
+
+$filter=[FirstName,LastName]%20in%20[["John","Doe"],["Jane","Smith"]]&
```
ralfhandl, line 978: I thought this content-type would encode space as `+`.

ralfhandl (commented):
Question on encoding of space

odata-url-conventions/4 Resource Path.md
HeikoTheissen: Indeed, it produces `%24filter=[FirstName,LastName]+in+[["John","Doe"],...`. Does OData accept this? Or can we demand that it accepts this in the case of `/$query`?

odata-url-conventions/4 Resource Path.md
```diff
@@ -979,18 +980,16 @@ Example ##ex: passing multiple system query options in the request body
...
-  <input name="$filter"
-    value='[FirstName,LastName] in [["John","Doe"],["Jane","Smith"]]'>
+  <input name="$filter" value="LastName eq 'P&G'">
   <input name="$select" value="FirstName,LastName">
 </form>
 ```
-Note `enctype="text/plain"` would not encode the spaces and would produce a newline
-instead of the `&` demanded by the OData syntax.
+Note `enctype="text/plain"` would not encode the spaces and ampersand and would
+separate the options with a newline instead of the `&` demanded by the OData syntax.
```
ralfhandl, line 994:
```suggestion
separate the options with a newline instead of the `&`.
```
I think the ampersand separator is required by `application/x-www-form-urlencoded`

odata-url-conventions/4 Resource Path.md
```diff
@@ -952,8 +952,9 @@ For content-type `text/plain`, it MUST use the same percent-encoding as in
 URLs (especially: no spaces, tabs, or line breaks allowed) and MUST
 follow the syntax rules described in chapter Query Options.
 For content-type `application/x-www-form-urlencoded`, it MUST first percent-decode
-the request body and then replace `+` with `%20`. The result MUST then be
-treated as the query portion of the URL like for content-type `text/plain`.
+the request body except for `%26` (ampersand) and `%3D` (equals) and then
+replace `+` with `%20`. The result MUST then be treated as the query portion
```
ralfhandl, line 956: Actually: decode `+` as space. And we have https://issues.oasis-open.org/browse/ODATA-1540 for allowing `+` as an encoding for space, so servers supporting that don't need special handling here.

ralfhandl (commented):
See comments

odata-url-conventions/4 Resource Path.md
ralfhandl: See https://issues.oasis-open.org/browse/ODATA-1540, where we want to entice servers to accept this (and discourage clients to use it 😄)

odata-url-conventions/4 Resource Path.md
HeikoTheissen: The OData syntax requires that system query options are separated by a `&`. (There cannot be newlines in a URL.)

ralfhandl: True, for URLs.

In requests bodies of content-type `application/x-www-form-urlencoded` it's the definition of that content type that requires the ampersands, just as you explain that `text/plain` requires newline as a separator.

Nothing OData-specific here.

HeikoTheissen: A `POST ~/$query` with `Content-Type: text/plain` requires an `&` as separator between the system query options, because the current spec says:
> The request body MUST use the content-type text/plain. It contains the query portion of the URL and MUST use the same percent-encoding as in URLs (especially: no spaces, tabs, or line breaks allowed) and MUST follow the syntax rules described in chapter Query Options.

This is OData-specific and does not align with the behavior of text/plain HTML forms.

ralfhandl: Too bad: can we also allow the default behavior of text/plain?

ralfhandl: But besides that, the x-www ampersand thingy has nothing to do with OData, so better remove that half sentence.

odata-url-conventions/4 Resource Path.md
```diff
@@ -946,10 +946,12 @@ Requests to paths ending in `/$query` MUST use the `POST` verb. Query
...
 request URL are processed together.
 
-The request body MUST use the content-type `text/plain`. It contains the
-query portion of the URL and MUST use the same percent-encoding as in
-URLs (especially: no spaces, tabs, or line breaks allowed) and MUST
-follow the syntax rules described in chapter Query Options.
+The request body MUST use the content-type `text/plain` or `application/x-www-form-urlencoded`.
+It contains the query portion of the URL.
+
+For content-type `text/plain`, the individual query options MUST be separated by `&` or newlines
```
HeikoTheissen, line 952: If we allow newlines as separator here, the reason for introducing `application/x-www-form-urlencoded` is gone. Shall we remove it?

ralfhandl (approved):

odata-url-conventions/4 Resource Path.md
```diff
@@ -946,12 +946,12 @@ Requests to paths ending in `/$query` MUST use the `POST` verb. Query
 options specified in the request body and query options specified in the
 request URL are processed together.
 
-The request body MUST use the content-type `text/plain` or `application/x-www-form-urlencoded`.
+The request body MUST use `Content-Type: text/plain` or `Content-Type: `application/x-www-form-urlencoded`.
```
ralfhandl, line 949:
```suggestion
The request body MUST use `Content-Type: text/plain` or `Content-Type: application/x-www-form-urlencoded`.
```

ralfhandl (approved):

ralfhandl (approved):

ralfhandl (dismissed):

odata-url-conventions/4 Resource Path.md
```diff
@@ -998,4 +1000,39 @@ Content-Type: text/plain
...
+      enctype="text/plain">
+  <input name="$filter"
+    value='[FirstName,LastName]%20in%20[["John","Doe"],["Jane","Smith"]]'>
+</form>
+```
+:::
+
+For content-type `application/x-www-form-urlencoded`, the individual query options MUST be separated by `&`
+and their names and values MAY be percent-encoded even for characters other than `%26` (ampersand) and
+`%3D` (equals).
```
ralfhandl, line 1015: Isn't that defined by `application/x-www-form-urlencoded`?

We need a normative reference and point to it here.

ralfhandl (changes requested):

odata-url-conventions/4 Resource Path.md
```diff
@@ -985,10 +985,12 @@ Requests to paths ending in `/$query` MUST use the `POST` verb. Query
...
-The request body MUST use `Content-Type: text/plain`. It contains the
-query portion of the URL and MUST use the same percent-encoding as in
-URLs (especially: no spaces, tabs, or line breaks allowed) and MUST
-follow the syntax rules described in chapter Query Options.
+The request body MUST use `Content-Type: text/plain` or `Content-Type: application/x-www-form-urlencoded`.
+It contains the query portion of the URL.
+
+For `Content-Type: text/plain`, the individual query options MUST be separated by `&` or newlines
+and MUST use the same percent-encoding as in URLs (especially: no spaces, tabs, or line breaks allowed)
+and MUST follow the syntax rules described in [chapter ##QueryOptions].
```
HeikoTheissen, line 993: This does not follow the normative [`text/plain` encoding algorithm](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#text/plain-encoding-algorithm), which allows only separation by newlines.

odata-url-conventions/4 Resource Path.md
```diff
@@ -998,4 +999,41 @@ Content-Type: text/plain
...
+    value='[FirstName,LastName]%20in%20[["John","Doe"],["Jane","Smith"]]'>
+</form>
+```
+:::
+
+For `Content-Type: application/x-www-form-urlencoded`, the request body MUST be
+the value of _output_ after running the
+[`application/x-www-form-urlencoded` serializer](https://url.spec.whatwg.org/#concept-urlencoded-serializer)
+in the [URL Living Standard](#_url), section 5.2, with _tuples_ being the list
+of name/value pairs for the individual query options.
```
HeikoTheissen, line 1016: Can we also use this reference to replace the second half of [OData-URL, section 2.1]?

ralfhandl: No, the WhatWG URL Spec is in parts contradicting RFC3986, which is why we spell out the "safe" compromise

> Note: neither [RFC3986](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part2-url-conventions/odata-v4.02-csd01-part2-url-conventions.html#rfc3986) nor this specification assign special meaning to “+” (octet 0x2B). Some implementations decode “+” (octet 0x2B) as space (octet 0x20), others take it literally.
> 
> Clients SHOULD percent-encode space (octet 0x20) as %20 and “+” (octet 0x2B) as %2B and avoid the ambiguous “+” (octet 0x2B) in URLs.

ralfhandl: We also added to protocol

> [12.3 Interoperable OData Clients](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#InteroperableODataClients)
> ...
> 10. MUST encode the plus character (octet `0x2B`) as `%2B` in URLs to avoid servers mis-interpreting the plus character as an encoded space

odata-url-conventions/4 Resource Path.md
ralfhandl: Do you want to open an issue for only allowing newline?

odata-url-conventions/Appendix.md
```diff
@@ -50,6 +50,10 @@ https://www.rfc-editor.org/info/rfc3986.
 _Leiba, B., "Ambiguity of Uppercase vs Lowercase in RFC 2119 Key Words", BCP 14, RFC 8174, DOI 10.17487/RFC8174, May 2017_.
 https://www.rfc-editor.org/info/rfc8174.
 
+###### <a name="_url">[URL]</a>
+_URL Living Standard._  
+https://url.spec.whatwg.org/.
```
ralfhandl, line 55: This WhatWG standard partially contradicts the IETF standard RFC3986, so I'd prefer a reference to an IETF RFC.

ralfhandl, line 55: Cross-check with OASIS citation rules for W3C specs

odata-url-conventions/4 Resource Path.md
```diff
@@ -998,4 +999,41 @@ Content-Type: text/plain
...
+
+%24filter=LastName+eq+%27P%26G%27&%24select=FirstName,LastName
+```
+This POST request would result from submitting the HTML form
+```html
+<form method="post" action="http://host/service/People/$query"
+      enctype="application/x-www-form-urlencoded">
+  <input name="$filter" value="LastName eq 'P&G'">
+  <input name="$select" value="FirstName,LastName">
+</form>
```
ralfhandl, line 1032: Heiko to test this example

odata-url-conventions/4 Resource Path.md
```diff
@@ -998,4 +999,41 @@ Content-Type: text/plain
...
+in the [URL Living Standard](#_url), section 5.2, with _tuples_ being the list
+of name/value pairs for the individual query options.
+
+::: example
+Example ##ex: passing multiple system query options in the request body
+```
+POST http://host/service/People/$query
+Content-Type: application/x-www-form-urlencoded
+
+%24filter=LastName+eq+%27P%26G%27&%24select=FirstName,LastName
```
ralfhandl, line 1024: `$` need not be encoded, the deserializer from WhatWG URL also accepts it.

Servers MUST use the deserializer from ...

Clients MUST serialize it so that the deserializer can make sense of it, for example by using the serializer described in the URL spec.

odata-url-conventions/Appendix.md
HeikoTheissen: https://www.w3.org/TR/html/ redirects to https://html.spec.whatwg.org/multipage/ and this contains no citation instructions.

odata-url-conventions/4 Resource Path.md
HeikoTheissen: @mikepizzo, your opinion is appreciated.

odata-url-conventions/4 Resource Path.md
ralfhandl, line 1: Keep 4.01 rules for text/plain unchanged and only add form-urlencoded

ralfhandl: Extend example for text/plain with a $orderby to show more than one parameter

Use same example for form-urlencoded to show the different encoding for spaces.

Optional: extend examples to point out other encoding differences (Ralf to provide link to encoding comparison).

ralfhandl: Summary (and brief history) of different percent-encoding standards: [OpenAPI 3.0.4 Draft, Appendix E: Percent-Encoding and Form Media Types](https://github.com/OAI/OpenAPI-Specification/blob/v3.0.4-dev/versions/3.0.4.md#appendix-e-percent-encoding-and-form-media-types)

ralfhandl (commented):
+1, added link to summary/differences of percent-encoding standards

ralfhandl (approved):

PR #181, opened 2023-09-29
Add slash if missing

HeikoTheissen:
`app.get("/:doc/", ...)` is no different from `app.get("/:doc", ...)`. Both variants also serve requests like `/odata-csdl?variant=json`, but then the styles and images cannot be loaded.

This PR redirects such a request to `/odata-csdl/?variant=json`.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -17,39 +17,45 @@ var app = express()
     var docs = [];
     iterator(function (srcname, name, variant) {
       docs.push(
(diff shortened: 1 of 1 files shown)


ralfhandl (approved):
Thanks, tried to solve it with `strict routing` and couldn't make it work. Would be more elegant because the redirect code would move to `.get("/:doc", ...`

HeikoTheissen:
The following would work:
```js
app.get(/\/$/, redirect)
   .get("/:doc", serve);
```

PR #183, opened 2023-09-29
Improvements to the web server

HeikoTheissen:
This PR serves to collect improvements to the web server:
* Don't use `iterator.js` during start-up so that new documents can be added while the server runs.
* Include SVG during build, not as images at runtime


ralfhandl (approved):

lib/number.js
```diff
@@ -137,16 +137,25 @@ class Number {
               var m1 = line.match(/^\$\$\$(.*?isec)\$\$\$$/);
               if (m1) this.tableofcontents(this.toc[m1[1]]?.sub || [], out, "");
               else {
+                m1 = line.match(/^\$\$\$include\s+(.*?)\$\$\$$/);
```
ralfhandl, line 140: No trailing whitespace allowed after the closing `$$$`? That could be hard to spot.

ralfhandl: If the images are embedded in the Markdown and HTML files, then we don't need to copy the `images` subfolder to `/docs`.

HeikoTheissen: Perhaps we also need *real* images later.

odata-temporal-ext/svg/api-1.drawio.svg
ralfhandl, line 1: I'd prefer `diagrams` as folder name to be consistent with Data Aggregation

lib/README.md
```diff
@@ -22,6 +22,7 @@ The [`number.js`](number.js) module generates a single Markdown document by prep
 - Generate section and example numbers
 - Generate a table of contents
 - Resolve references
+- Include files like `$$$include images/drawing.svg$$$`
```
ralfhandl, line 25:
```suggestion
- Include files like `$$$include diagrams/drawing.svg$$$`
```

ralfhandl, line 25:
```suggestion
- Copy `images` subfolder to output folder
- Include files like `$$$include diagrams/drawing.svg$$$`
```

ralfhandl (approved):

PR #189, opened 2023-10-19
ODATA-1446

HeikoTheissen:


odata-protocol/11.5 Operations.md
```diff
@@ -208,6 +208,10 @@ POST http://host/service/MyShoppingCart()/Items
 ```
 :::
 
+If the function returns a value of type `Edm.Stream` and no additional path
+segments follow the function invocation, the response to the `GET` request
+obeys the rules for [requesting stream properties](#RequestingStreamProperties).
```
ralfhandl, line 213:
```suggestion
follows the rules for [requesting stream properties](#RequestingStreamProperties).
```

odata-protocol/11.5 Operations.md
```diff
@@ -208,6 +208,10 @@ POST http://host/service/MyShoppingCart()/Items
...
+
```
ralfhandl, line 214: What happens if there are additional path segments after the function invocation? Do we allow anything other than an invocation of another function with binding parameter type Edm.Stream?

ralfhandl (commented):
+1, only two minor comments

odata-protocol/11.5 Operations.md
HeikoTheissen: I cannot imagine anything other than that. But what happens then is explained elsewhere (I hope).

odata-protocol/11.5 Operations.md
HeikoTheissen: I wanted to avoid using the word "follow" twice with different meanings in the same sentence.

chrisspre: I think it should be neither `obeys` nor `follows` since that would imply that that section is talking about function.
I would say something along the lines of "the response to the GET request is analogous to the rules for ... "

mikepizzo (approved):

PR #191, opened 2023-10-20
ODATA-1370: disambiguate action/function and overload

ralfhandl:
https://issues.oasis-open.org/browse/ODATA-1370: distinguish between

- an action/function as the named set of all its signature definitions aka overloads
- a single signature definition aka overload

CSDL XML doesn't have a direct representation of an action/function, the elements `edm:Action` and `edm:Function` represent a single overload.


odata-csdl/12 Action and Function.md
```diff
@@ -285,8 +285,8 @@ of the collection and specifies whether the collection MAY contain
 `null` values.
 
 For single-valued return types the value `true` means that the action or
-function MAY return a single `null` value. The value `false` means that
-the action or function will never return a `null` value and instead will
+function overload MAY return a single `null` value. The value `false` means that
+the action or function overload will never return a `null` value and instead will
```
HeikoTheissen, line 289: Make analogous change for the XML variant.
Must all function overloads with the same name have the same return type nullability?

HeikoTheissen (changes requested):

odata-csdl/12 Action and Function.md
ralfhandl: I assumed that the full return type (including nullability and type facets) must be identical. Added this above.

HeikoTheissen (approved):

HeikoTheissen (approved):

odata-csdl/12 Action and Function.md
```diff
@@ -79,38 +78,39 @@ MUST be unique within its schema.
...
-- All unbound functions with the same
-function name within a schema MUST specify the same return type.
-
-An unbound function MAY have the same name as a bound function.
+- All unbound function overloads with the same
+function name within a schema MUST specify
+the same [return type](#ReturnType) with identical nullability and type facets.
 
 Note that [type definitions](#TypeDefinition) can be used to
 disambiguate overloads for both bound and unbound functions, even if
```
HeikoTheissen, line 116: 1. Why "both bound and unbound"? Unbound function overloads can be distinguished by function name and parameter names, without looking at types.
2. unbound functions -> unbound function overloads
3. Does this also work for actions?

HeikoTheissen (changes requested):

odata-csdl/12 Action and Function.md
ralfhandl: The second rule "The combination of function name and ordered set of parameter types MUST be unique within a schema" ignores the parameter _names_, so overloads
~~~
function forecastFor(year: Edm.String)
function forecastFor(quarter: Edm.String)
~~~
would not be allowed as both have the same type. What would be allowed is
~~~
function forecastFor(year: My.Year)
function forecastFor(quarter: My.Quarter)

typedef Year = Edm.String
typedef Quarter = Edm.String
~~~

This is not necessary for actions as overload resolution is based exclusively on the binding parameter _type_.

HeikoTheissen: What's the problem with differently named parameters that have the same type? The parameter names always appear in the URL.

ralfhandl: @mikepizzo Was this a restriction in the .NET platform?

HeikoTheissen (approved):

HeikoTheissen (approved):

mikepizzo:
The current proposal groups all overloads with the same name as a single action/function.  This matches how represented in JSON. However, our semantic rules for functions are around overloads with the same name and binding parameter (i.e., overloads with the same name and binding parameter must have the same return type).

@mikepizzo to attempt a PR to explore defining an action/function as same name/binding parameter.

mikepizzo (changes requested):
The current proposal groups all overloads with the same name as a single action/function. This matches how represented in JSON. However, our semantic rules for functions are around overloads with the same name and binding parameter (i.e., overloads with the same name and binding parameter must have the same return type).

Would like to explore what this would look like if we defined an action/function as the set of overloads with same name and binding parameter.

odata-csdl/12 Action and Function.md
mikepizzo: This was to support languages that resolved overloads by parameter types.

PR #197, opened 2023-10-23
ODATA-1593

ralfhandl:


odata-json-format/15 Delta Payload.md
```diff
@@ -478,17 +492,152 @@ inline as a delta
...
+  3. Delete customer 'ANTON' - failed
+  4. Change customer 'ALFKI':
+     1. Create order 11011 - succeeded, not mentioned in response
+     2. Add link to existing order 10692 - succeeded, not mentioned in response
+     3. Change `RequiredDate` of related order 10835 - failed
+     4. Delete link to order 10643 - succeeded, not mentioned in response
+  5. Add link between customer 'ANATR' and order 10643 - failed without further info
+  6. Delete link between customer 'DUMON' and order 10311 - failed without further info
+```json
+HTTP/1.1 200 Ok
```
HeikoTheissen, line 532:
```suggestion
HTTP/1.1 200 OK
```

odata-json-format/15 Delta Payload.md
```diff
@@ -478,17 +492,152 @@ inline as a delta
...
+HTTP/1.1 400 Bad Request
+Content-Type: application/json
+Content-Length: ###
+
+{
+  "error": {
+    "code": "incmplt",
+    "message": "Required field(s) not provided",
+    "target": "Customers('EASTC')/Address",
+    "com.sap.vocabularies.Common.v1.additionalTargets": [ "Customers('EASTC')/Industry", "Customers('EASTC')/VATRegistration" ]
```
HeikoTheissen, line 639: Do we want this SAP annotation in the OASIS spec?

ralfhandl: Yes, explicitly requested in issue https://issues.oasis-open.org/browse/ODATA-1593

odata-json-format/15 Delta Payload.md
```diff
@@ -478,17 +492,152 @@ inline as a delta
...
 }
 ```
+
+Assuming all changes can be applied without errors, the response would be
+```
+HTTP/1.1 204 No Content
+Preference-Applied: return=minimal, continue-on-error
+
+
+
```
HeikoTheissen, line 517: Why three blank lines? My server produces only one.

HeikoTheissen (changes requested):

odata-json-format/15 Delta Payload.md
ralfhandl: Actually there have to be two blank lines because the header block of an HTTP message is ended by LF LF. My server and browser shows this when selecting to highlight:
image

Which are exactly the two empty lines I want.

HeikoTheissen: My web server produces the following:
```
HTTP/1.1 204 No Content <CRLF>
Preference-Applied: return=minimal, continue-on-error<CRLF>
<CRLF>
```
The final two <CRLF>s are at the end of the second and third line. There is no fourth line. Does it look the same for you?

odata-json-format/15 Delta Payload.md
```diff
@@ -478,17 +492,152 @@ inline as a delta
...
+Content-Length: ###
+Preference-Applied: return=minimal, continue-on-error
+
+{
+  "@context": "#$delta",
+  "value": [
+    {
+      "@Org.OData.Core.V1.ContentID": "1",
+      "CustomerID": "EASTC",
+      "@removed": {},
```
HeikoTheissen, line 543: Does `@removed` mean "not created" in this case?

ralfhandl: Seems so. We decided to "invert" the response: inserts become deletes, deletes become inserts, updates remain updates.

odata-json-format/15 Delta Payload.md
ralfhandl: The produced HTML looks fine for me:
https://github.com/oasis-tcs/odata-specs/blob/09fb4d4b02c452e58758784047e8421a6643b9d1/docs/odata-json-format/odata-json-format.html#L1397-L1401

Line 1399 is the separator line between the header block and the response body, and line 1400 is the empty response body

ralfhandl: On the wire it would be `...continue-on-error<LF><LF>`.

HeikoTheissen: The code on the wire produces _one_ empty line (the one between the two `<LF>`s). But the generated HTML contains _two_ empty lines.

ralfhandl: That's a tricky visualization question.

Assume the response is an empty JSON object, on the wire `last: header<LF><LF>{}`. This would be visualized as
~~~http
last: header

{}
~~~

If the response has an empty response body, we'd get on the wire `last: header<LF><LF>` on the wire, just the `{}` removed. So if we remove the curlies also from the visualization, we get
~~~http
last: header


~~~

In both cases the visualization has actually three <LF> octets whereas the message only has two - there is no trailing <LF> required at the end of an HTTP message.

HeikoTheissen (approved):

HeikoTheissen (dismissed):

HeikoTheissen (approved):

PR #201, opened 2023-10-26
ODATA-1604

ralfhandl:
Resolves https://issues.oasis-open.org/browse/ODATA-1604

### DIFF
--- a/odata-url-conventions/5 Query Options.md
+++ b/odata-url-conventions/5 Query Options.md
@@ -2316,6 +2316,7 @@ Search expressions are used within the
 entities matching the specified expression.
 
 *Terms* can be any single word to be matched within the expression.
+Leading and trailing spaces are not considered part of a word.
 
 Terms enclosed in double-quotes comprise a *phrase*.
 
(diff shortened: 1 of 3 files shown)


odata-url-conventions/5 Query Options.md
```diff
@@ -2316,6 +2316,7 @@ Search expressions are used within the
 entities matching the specified expression.
 
 *Terms* can be any single word to be matched within the expression.
+Leading and trailing spaces are not considered part of a word.
```
HeikoTheissen, line 2319: That is already implied by my understanding of what a "word" is.

odata-url-conventions/5 Query Options.md
```diff
@@ -2316,6 +2316,7 @@ Search expressions are used within the
 entities matching the specified expression.
```
HeikoTheissen, line 2316:
```suggestion
entities matching the specified expression. Leading and trailing spaces are not considered part of the search expression.
```
Covers cases like `$search=%20"%20"%20-%20` which searches for space AND hyphen.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

PR #207, opened 2023-10-31
Typo: Principal property

HeikoTheissen:
The terms "principal property" and "principal entity" are used but where are they defined?


ralfhandl:
They are defined in [CSDL 8.5 Referential Constraint](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_ReferentialConstraint)

ralfhandl:
Just noticed the same spelling error in the CSDL specs 😟

PR #216, opened 2023-11-27
`$Apply` and `$Function` in Appendix B

HeikoTheissen:


ralfhandl (approved):
Which problem does this solve?

HeikoTheissen:
> Which problem does this solve?

`$Apply` was listed as an annotation member in appendix B, but `$Function` was not. The two occur together (like `$Type` and `$Collection`).

PR #217, opened 2023-11-30
ODATA-1559

ralfhandl:


odata-protocol/1 Introduction.md
```diff
@@ -287,10 +287,12 @@ one or both of them may differ from convention.
 
 ## ##subsec Transient Entities
 
-Transient entities are instances of an entity type that are "calculated
+Transient entities are derived from instances of an entity type that are "calculated
 on the fly" and only exist within a single payload. They cannot be
```
HeikoTheissen, line 291: Are "calculation on the fly" and "derivation" two steps? Or is this redundant?
```suggestion
Transient entities are instances of an entity type that are derived from entities of the same entity type and only exist within a single payload. They cannot be
```

odata-protocol/1 Introduction.md
```diff
@@ -287,10 +287,12 @@ one or both of them may differ from convention.
...
-reread or updated and consequently possess neither a stable entity-id
-nor a read URL or an update URL.
+updated and consequently possess neither a stable entity-id
+nor an update URL.
+A transient entity may have a read URL, which may be used to re-derive
+the transient entity from the current state of the underlying entity.
```
HeikoTheissen, line 295:
```suggestion
the transient entity from the current state of the underlying entity or entities.
```

HeikoTheissen (approved):

GEricson:
Ralf,
A possible simplification in green.

<p>Transient entities are instances of an entity type that are “calculated on the fly” and only exist within a single payload. They cannot be reread or updated and consequently possess neither a stable entity-id nor a read URL or an update URL.</p>
<p>Transient entities are instances of an entity type that are derived from entities of the same entity type and only exist within a single payload. They cannot be updated and consequently possess neither a stable entity-id nor an update URL. A transient entity may have a read URL, which may be used to re-derive the transient entity from the current state of the underlying entity or entities.</p>
<p>A transient entity is a representation of an entity and only exists within a single payload. A transient entity cannot be updated and consequently does not possess a stable entity-id or an update URL. A transient entity may have a read URL, which may be used to re-derive the transient entity from the entity that it represents.</p>

George




Internal Use - Confidential

ralfhandl:
@GEricson 

> A possible simplification in green

I don't see any green.

Two remarks: 
- email replies to Github notifications end up with horrible formatting and are better avoided, please use the web page to comment
- please use the "Add a suggestion" feature of Github pull requests
  - hover over the line you want to comment on and press the blue "+" button at the beginning of the line
    image
  - then press the "Add a suggestion" button
    image

docs/odata-protocol/odata-protocol.html
```diff
@@ -590,7 +590,7 @@ <h2 id="42-read-urls-and-edit-urls"><a name="ReadURLsandEditURLs" href="#ReadURL
...
 <h2 id="43-transient-entities"><a name="TransientEntities" href="#TransientEntities">4.3 Transient Entities</a></h2>
-<p>Transient entities are instances of an entity type that are “calculated on the fly” and only exist within a single payload. They cannot be reread or updated and consequently possess neither a stable entity-id nor a read URL or an update URL.</p>
+<p>Transient entities are instances of an entity type that are derived from entities of the same entity type and only exist within a single payload. They cannot be updated and consequently possess neither a stable entity-id nor an update URL. A transient entity may have a read URL, which may be used to re-derive the transient entity from the current state of the underlying entity or entities.</p>
```
chrisspre, line 593: "derived from entities" I have two comments 
1) are we consistent in using "entity" and "**instance** of an entity type " as essentially synonymous?
2) the phrase "derived from" is vague in this context. In the meeting it was said that it means subtype but I do not think this is consistent since a "derived entity" can be so fundamentally different that the derived instance can not be used in all operations where the source instance can be used (the definition of subtype) . A problem that exits in structurally typed programming languages as well. 

I would rather see something that in spirit is like "calculated on the fly" something like "created by a transformation of a source entity"  
In API design discussion we sometimes use the phrase ephemeral for such an entity.

my proposal: 

`<p>Transient entities are instances of an entity type that are the result of transforming an entity of that type and only exist within a single payload. They cannot be updated and consequently possess neither a stable entity-id nor an update URL. A transient entity may have a read URL, which may be used to apply the same transformation to  the current state of the underlying entity or entities.</p>`

GeraldKrause: I like this proposal and would go one step further and not require that the transient entity must be "calculated", "transformed" or "derived" from an entity of that type. I do not see any reason why it could not be created freely in any other way.

Here's my revised proposal (only removed "are the result of transforming an entity of that type and"):

`<p>Transient entities are instances of an entity type that only exist within a single payload. They cannot be updated and consequently possess neither a stable entity-id nor an update URL. A transient entity may have a read URL, which may be used to apply the same transformation to  the current state of the underlying entity or entities.</p>`

ralfhandl: > 1. are we consistent in using "entity" and "**instance** of an entity type " as essentially synonymous?

Yes, entities are instances of an entity type.

ralfhandl: Removing "transforming" from the first sentence makes it hard to understand "apply the same transformation" in the last sentence. I'd stick with Christof's proposal.

ralfhandl: Applied Christof's proposal, with the modification of allowing multiple underlying entities of the same entity type.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

odata-protocol/1 Introduction.md
```diff
@@ -294,11 +294,9 @@ one or both of them may differ from convention.
 ## ##subsec Transient Entities
 
 Transient entities are instances of an entity type that are 
-the result of transforming entities of that entity type
-and only exist within a single payload.
+dynamically generated on request and only exist within a response payload.
 They cannot be updated and consequently possess neither a stable entity-id nor an update URL.
-A transient entity may have a read URL, which may be used to apply the same transformation to
-the current state of the underlying entity or entities.
+A transient entity may have a read URL, which generates a new transient entity using the same algorithm.
```
HeikoTheissen, line 299:
```suggestion
A transient entity may have a read URL, which regenerates the transient entity using the same algorithm.
```
I prefer "regenerates" over "generates a new", because the read URL remains the same.

HeikoTheissen (changes requested):

odata-protocol/1 Introduction.md
ralfhandl: The URL remains the same, and potentially returns a different transient entity.

HeikoTheissen (approved):

mikepizzo (approved):

PR #221, opened 2023-12-01
Analyse with puppeteer

HeikoTheissen:
A tool that helps with the analysis of the generated HTML documents. For example, syntax errors in code snippets can be detected with
```
node lib/selector.mjs ".json .er" "self::*[.!='…' and not(starts-with(.,' '))]/text()"
```
The first argument is a CSS selector, the (optional) second is an XPath expression.

### DIFF
--- a/lib/README.md
+++ b/lib/README.md
@@ -96,6 +96,10 @@ The following scripts can be executed manually or as part of a GitHub Action:
   - a copy of the common [`styles`](../styles) folder
   - a copy of the document-specific `*/images` folder, if this exists.
 - [`npm run pdf`](build-pdf.mjs) runs the PDF conversion and writes the PDF document into the [`docs/*`](../docs) folder.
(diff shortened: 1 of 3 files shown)


lib/README.md
```diff
@@ -96,6 +96,10 @@ The following scripts can be executed manually or as part of a GitHub Action:
   - a copy of the common [`styles`](../styles) folder
   - a copy of the document-specific `*/images` folder, if this exists.
 - [`npm run pdf`](build-pdf.mjs) runs the PDF conversion and writes the PDF document into the [`docs/*`](../docs) folder.
+- [`npm run select <CSS selector> [<XPath expression>]`](selector.mjs) selects parts of the generated HTML documents by executing a CSS selector and optionally an XPath expression relative to each match. For example, syntax errors in JSON code snippets can be detected with
+  ```sh
+  npm run select ".json .er" "self::*[.!='…']/text()"
+  ```
```
ralfhandl, line 102: Why not have a `postbuild` script that calls this validation?

ralfhandl (commented):
I think this is useful enough to be part of our build process

ralfhandl (approved):

PR #233, opened 2024-01-11
Bound actions and function overloads via container

HeikoTheissen:
Shall we also allow schema-bound paths like
```
MySchema.MyEntityType/MyNavigationProperty/MyFunction
```
?

### DIFF
--- a/odata-csdl/14 Vocabulary and Annotation.md
+++ b/odata-csdl/14 Vocabulary and Annotation.md
@@ -580,6 +580,16 @@ qualified name of action followed by parentheses containing the binding paramete
 all overloads of an [Action](#Action)| 
 qualified name of action| 
 <pre>`MySchema.MyAction`</pre>
+bound [Action](#Action) overload via container| 
+Navigation Property via container or Property via container, followed by bound Action overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty` 
+<br>` /MySchema.MyAction(Collection(MySchema.MyBindingType))`</pre>
+all bound [Action](#Action) overloads with given binding type via container| 
+Navigation Property via container or Property via container, followed by all overloads of an Action| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty` 
+<br>` /MySchema.MyAction`</pre>
 [Action Import](#ActionImport)| 
 qualified name of entity container followed by a segment containing the action import name| 
 <pre>`MySchema.MyEntityContainer/MyActionImport`</pre>
@@ -614,6 +624,16 @@ qualified name of function followed by parentheses containing the comma-separate
 all overloads of a [Function](#Function)| 
 qualified name of function| 
 <pre>`MySchema.MyFunction`</pre>
+bound [Function](#Function) overload via container| 
+Navigation Property via container or Property via container, followed by bound Function overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyComplexProperty` 
+<br>` /MySchema.MyFunction(MySchema.MyBindingType)`</pre>
+all bound [Function](#Function) overloads of given binding type via container| 
+Navigation Property via container or Property via container, followed by all overloads of a Function| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyComplexProperty` 
+<br>` /MySchema.MyFunction`</pre>
 [Function Import](#FunctionImport)| 
 qualified name of entity container followed by a segment containing the function import name| 
 <pre>`MySchema.MyEntityContainer/MyFunctionImport`
@@ -673,6 +693,15 @@ External targeting is possible for properties and navigation
 properties of singletons or entities in a particular entity set. These
 annotations override annotations on the properties or navigation
 properties targeted via the declaring structured type.
+
+External targeting is also possible for action and function overloads that
+are bound to properties or navigation properties of singletons or entities in
+a particular entity set. These annotations override annotations targeting the
+action or function overloads directly.
+
+Note "all bound Action or Function overloads of given binding type via
+container" references all overloads where the binding parameter is an instance
+or a collection of the given type.
 
 ## ##subsec Constant Expression
 
--- a/docs/odata-csdl-json/odata-csdl-json.html
+++ b/docs/odata-csdl-json/odata-csdl-json.html
@@ -2276,6 +2276,16 @@ Allowed values: -1.234567e3, 1e-101, 9.999999e96, not allowed values: 1e-102 and
 <td><pre><code>MySchema.MyAction</code></pre></td>
 </tr>
 <tr class="odd">
+<td>bound <a href="#Action">Action</a> overload via container</td>
+<td>Navigation Property via container or Property via container, followed by bound Action overload</td>
+<td><pre><code>MySchema.MyEntityContainer/MyEntitySet</code> <br><code> /MyNavigationProperty</code> <br><code> /MySchema.MyAction(Collection(MySchema.MyBindingType))</code></pre></td>
+</tr>
+<tr class="even">
+<td>all bound <a href="#Action">Action</a> overloads with given binding type via container</td>
+<td>Navigation Property via container or Property via container, followed by all overloads of an Action</td>
+<td><pre><code>MySchema.MyEntityContainer/MyEntitySet</code> <br><code> /MyNavigationProperty</code> <br><code> /MySchema.MyAction</code></pre></td>
+</tr>
+<tr class="odd">
 <td><a href="#ActionImport">Action Import</a></td>
 <td>qualified name of entity container followed by a segment containing the action import name</td>
 <td><pre><code>MySchema.MyEntityContainer/MyActionImport</code></pre></td>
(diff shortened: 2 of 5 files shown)


odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -580,6 +580,16 @@ qualified name of action followed by parentheses containing the binding paramete
 all overloads of an [Action](#Action)| 
 qualified name of action| 
 <pre>`MySchema.MyAction`</pre>
+bound [Action](#Action) overload via container| 
+Navigation Property via container or Property via container, followed by bound Action overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty`
```
ralfhandl, line 586: What about paths without a navigation property, to differentiate actions on multiple entity sets sharing the same entity type?

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -580,6 +580,16 @@ qualified name of action followed by parentheses containing the binding paramete
...
+bound [Action](#Action) overload via container| 
+Navigation Property via container or Property via container, followed by bound Action overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty` 
+<br>` /MySchema.MyAction(Collection(MySchema.MyBindingType))`</pre>
+all bound [Action](#Action) overloads with given binding type via container| 
+Navigation Property via container or Property via container, followed by all overloads of an Action| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty` 
+<br>` /MySchema.MyAction`</pre>
```
ralfhandl, line 592:
```suggestion
<br>` /MySchema.MyAction(MySchema.MyBindingType)`</pre>
```
to be symmetric with the collection case above.

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -614,6 +624,16 @@ qualified name of function followed by parentheses containing the comma-separate
 all overloads of a [Function](#Function)| 
 qualified name of function| 
 <pre>`MySchema.MyFunction`</pre>
+bound [Function](#Function) overload via container| 
+Navigation Property via container or Property via container, followed by bound Function overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyComplexProperty`
```
ralfhandl, line 630: What about paths without a navigation property, to differentiate functions on multiple entity sets sharing the same entity type?

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -614,6 +624,16 @@ qualified name of function followed by parentheses containing the comma-separate
 all overloads of a [Function](#Function)| 
 qualified name of function| 
 <pre>`MySchema.MyFunction`</pre>
+bound [Function](#Function) overload via container| 
+Navigation Property via container or Property via container, followed by bound Function overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyComplexProperty` 
+<br>` /MySchema.MyFunction(MySchema.MyBindingType)`</pre>
```
ralfhandl, line 631: Why the binding type? Text says "all overloads", so I'd expect
```suggestion
<br>` /MySchema.MyFunction`</pre>
```

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -580,6 +580,16 @@ qualified name of action followed by parentheses containing the binding paramete
 all overloads of an [Action](#Action)| 
 qualified name of action| 
 <pre>`MySchema.MyAction`</pre>
+bound [Action](#Action) overload via container| 
+Navigation Property via container or Property via container, followed by bound Action overload| 
+<pre>`MySchema.MyEntityContainer/MyEntitySet` 
+<br>` /MyNavigationProperty` 
+<br>` /MySchema.MyAction(Collection(MySchema.MyBindingType))`</pre>
```
ralfhandl, line 587: Please add second example without `Collection(...)`
```suggestion
<pre>`MySchema.MyEntityContainer/MyEntitySet`
<br>` /MySchema.MyAction(MySchema.MyBindingType)`</pre>
```

ralfhandl (changes requested):
Please also consider the case via container on entity set or singleton without an additional property or navigation property before the action/function.

Also consider multi-segment paths for nested containment/structures

HeikoTheissen:
> Please also consider the case via container on entity set or singleton without an additional property or navigation property before the action/function.
> 
> Also consider multi-segment paths for nested containment/structures

We don't have all these examples for the other cases either. Let's discuss this in the TC.

mikepizzo:
> Shall we also allow schema-bound paths like
> 
> ```
> MySchema.MyEntityType/MyNavigationProperty/MyFunction
> ```
> 
> ?

Yes.

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -676,26 +676,27 @@ properties targeted via the declaring structured type.
 
 External targeting is also possible for action and function overloads that
 are bound to a certain type of model element. The target path then consists of
-two path expressions separated by a forward slash. The first path expression
-references one of the following model elements:
+two path expressions from the table above separated by a forward slash. The
+first path expression references one of the following model elements:
 - Navigation Property via container
```
ralfhandl, line 681:
```suggestion
- Entity set or singleton
- Navigation Property via container
```

ralfhandl (changes requested):

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -676,26 +676,27 @@ properties targeted via the declaring structured type.
...
 - Function overload
-- all overloads of a Function
 
 where the Action or Function overload is bound with binding parameter type
-equal to the type of the first path expression. These annotations override
-annotations targeting the action or function overloads directly.
+equal to the type of the first path expression, or the second path expression
+references
+- all overloads of an Action or
+- all overloads of a Function
```
ralfhandl, line 694: "all ... with a binding type equal to or derived from the type of the model element referenced by the first path expression"

HeikoTheissen: This restriction comes after the bulleted list.

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -674,6 +674,39 @@ properties of singletons or entities in a particular entity set. These
...
+Model element|means the binding parameter must
+-------------|--------------------------------
+Entity Set|lie in the entity set
+Singleton|be the singleton
+Navigation Property via container|be addressed via the path
+Property via container|be addressed via the path
+Navigation Property via structured type|be addressed via this structured type
+Property via structured type|be addressed via this structured type
+Complex Type|have the type or a subtype thereof
+Entity Type|have the type or a subtype thereof
```
HeikoTheissen, line 692:
```
MySchema.MyEntityType/MyAction(MyEntityType)
```
is equivalent to `MyAction(MyEntityType)` but
```
MySchema.MyEntityType/MyAction
```
has a meaning of its own. Is it OK to allow both?

HeikoTheissen:
What about targets like
```
MySchema.MyAction/$ReturnType/PropertyOfMyComplexReturnType
```
?

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -674,6 +674,44 @@ properties of singletons or entities in a particular entity set. These
...
+Navigation Property via structured type|addressed by a path that traverses the structured type
+Property via structured type|addressed by a path that traverses the structured type
+Complex Type|an instance or collection of the type or a subtype thereof
+Entity Type|an instance or collection of the type or a subtype thereof
+
+A binding parameter is "addressed via a resource path" if the path expression addressing the
+binding parameter starts with the resource path, it may then continue with segments
+described in [OData-URL](#ODataURL), sections 4.6 through 4.12. If the resource path
+addresses a collection, the binding parameter can also be a single instance of that
+collection if a segment described in [OData-URL](#ODataURL), sections 4.9 and 4.10, follows.
```
HeikoTheissen, line 699: Does the annotation target
```
MySchema.MyContainer/MyEntitySet/MyAction
```
address the bound action
```
POST MyEntitySet/MyComposableFunction()/MyAction
```
even if the composable function returns a _different_ entity set?

ralfhandl: I would expect that it does _not_, instead I'd expect that annotations for target
`MySchema.MyContainer/MyOtherEntitySet/MySchema.MyAction` apply if function `MySchema.MyComposableFunction` returns entities from `MyOtherEntitySet`.

HeikoTheissen: > instead I'd expect that annotations for target
> `MySchema.MyContainer/MyOtherEntitySet/MySchema.MyAction` apply if function `MySchema.MyComposableFunction` returns entities from `MyOtherEntitySet`.

Which is the case according to the first table row:
Model element|means the binding parameter must be
-------------|-----------------------------------
Entity Set|an instance or collection of instances from the entity set

HeikoTheissen: But the target
```
MySchema.MyContainer/MyOtherEntitySet/MyContainmentNavigationProperty/MySchema.MyAction
```
would not apply if the composable function returned entities from the contained entity set.

Must we distinguish between containment and non-containment navigation properties? Currently we have only

Model element|means the binding parameter must be
-------------|-----------------------------------
Navigation Property via container|addressed via a resource path to the navigation property
Navigation Property via structured type|addressed via a resource path to the navigation property that traverses an instance or collection of the structured type

ralfhandl: If the function returns totally unrelated entities, we can't express that (yet) in $metadata.

If the function returns related entities and is defined with
~~~xml
EntitySetPath="__it/_otherEntity/MyContainmentNavigationProperty"
~~~
then the client could figure out look for annotations with target
~~~
MySchema.MyContainer/MyOtherEntitySet/MyContainmentNavigationProperty/MySchema.MyAction
~~~

That is asking a lot, and I feel we are down a very exotic rabbit hole.

PR #246, opened 2024-02-09
ODATA-1637

ralfhandl:


HeikoTheissen (approved):

HeikoTheissen:
We do not state explicitly that `EntitySetPath` and `EntitySet` are unnecessary, right?

PR #250, opened 2024-02-22
ODATA-1642 - Correct treatment of key property values

ralfhandl:


odata-protocol/11.4 Data Modification.md
```diff
@@ -561,9 +561,7 @@ supporting this SHOULD advertise it by annotating the singleton with the
 term `Capabilities.UpdateRestrictions` (nested property `Upsertable`
 with value `true`) defined in [OData-VocCap](#ODataVocCap).
 
-Key and other non-updatable properties, as well as dependent properties
-that are not tied to key properties of the principal entity, MUST be
```
HeikoTheissen, line 565: What was meant by "tied to key properties of the principal entity"? Before we remove that statement, I would like to understand it.

ralfhandl: The removed statement was an _incomplete_ copy of the statement in https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_CreateanEntity

> properties that are tied to properties of the principal entity by a referential constraint [...] can be omitted and MUST be ignored if included in the request

Means if `Order` has key `OrderID` and an association to many `Item`, `Item` has two-part key `OrderID`,`ItemNo`, and there is a referential constraint between the two `OrderID` properties, then in `POST Orders(OrderID=4711)/_Items` I can omit the `OrderID` from the request body because its value is already defined by the request URL and the referential constraint. If I don't omit it, the server MUST ignore the value in the request body because the URL plus the referential constraint win.

HeikoTheissen: Confusion: The original statement says "are tied", the copied statement (to be deleted) says "are not tied".

ralfhandl:
Replaced by #1924

PR #1917, opened 2024-04-04
More rigorous definition of request references

HeikoTheissen:


odata-protocol/11.7 Batch Requests.md
```diff
@@ -140,11 +142,14 @@ identifier prefixed with a `$` character as the unquoted value of the
...
-references consist of a `$` character, followed by the request
-identifier of the preceding request, and optionally followed by a valid
-OData path.
+Services MAY support using values from a response body in the query part of
+the URL or in the request body of subsequent requests. Value references
+consist of a `$` character, followed by the request identifier of the preceding
+request. They evaluate to the referenced value, that is the value represented
+by the response body of that preceding request. If that value is a collection
+or has an entity or complex type, the value reference MAY be followed by a
+forward slash and a path expression that is evaluated relative to the
```
ralfhandl, line 151: No forward slash if it is a collection of entities and follow it with a paren-style key.

odata-protocol/11.7 Batch Requests.md
```diff
@@ -105,7 +105,9 @@ the rules for which individual requests require an identifier.
 Entities created by an [insert](#CreateanEntity) request or an [action](#InvokinganAction) can be
 referenced in the request URL of subsequent requests by using the
 request identifier prefixed with a `$` character as the first segment of
-the request URL. If the [`Location`](#HeaderLocation) header in the response contains a relative URL,
+the request URL. Services MUST treat this segment like the edit URL or read URL in the
```
ralfhandl, line 108: Don't mention read and edit URL

odata-protocol/11.7 Batch Requests.md
```diff
@@ -140,11 +142,14 @@ identifier prefixed with a `$` character as the unquoted value of the
...
 ### ##subsubsec Referencing Values from Response Bodies
 
-Services MAY support using values from a response body in the query part
-of the URL or in the request body of subsequent requests. Value
-references consist of a `$` character, followed by the request
-identifier of the preceding request, and optionally followed by a valid
-OData path.
+Services MAY support using values from a response body in the query part of
+the URL or in the request body of subsequent requests. Value references
+consist of a `$` character, followed by the request identifier of the preceding
```
ralfhandl, line 147:
```suggestion
consist of a `$` character, followed by the identifier of the preceding request
```

ralfhandl (dismissed):

odata-protocol/11.7 Batch Requests.md
```diff
@@ -105,7 +105,9 @@ the rules for which individual requests require an identifier.
 Entities created by an [insert](#CreateanEntity) request or an [action](#InvokinganAction) can be
 referenced in the request URL of subsequent requests by using the
 request identifier prefixed with a `$` character as the first segment of
-the request URL. If the [`Location`](#HeaderLocation) header in the response contains a relative URL,
+the request URL. Services MUST treat this segment like the URL in the
+[`Location`](#HeaderLocation) header of the response to the request identified by the segment.
+If the `Location` header in the response to the subsequent request contains a relative URL,
```
ralfhandl, line 110: An example would be helpful, showing Location headers in both responses

ralfhandl:
Homework: cross-check our resolution rules against https://www.rfc-editor.org/rfc/rfc3986#section-5, especially for multi-part responses.

HeikoTheissen:
Here's my argumentation:

"5.1.2. Base URI from the Encapsulating Entity" takes precedence over "5.1.3. Base URI from the Retrieval URI". In the case of an individual response within a $batch response the encapsulating entity is the $batch response, but this has no base URL for lack of a `Content-Location` header. Therefore 5.1.2 is not applicable in this case and 5.1.3 becomes relevant.

We should state in [OData-Protocol, section 11.7] that the "URI used to retrieve the representation" of an individual response within a $batch response is the corresponding individual request URL, in both multipart and JSON $batch.

odata-protocol/11.7 Batch Requests.md
```diff
@@ -13,7 +13,10 @@ service root.
 
 Individual requests within a batch request are evaluated according to
 the same semantics used when the request appears outside the context of
-a batch request.
+a batch request. The base URI of the response to one of the individual requests is
+determined according to [RFC3986, section 5.1.3](#rfc3986) with the
+"URI used to retrieve the representation" being the URL of the individual request
+(see [example ##batchcontentid]).
```
ralfhandl, line 19: Do we want a similar clarification in JSON Format?

ralfhandl (commented):
+1, one minor nit

ralfhandl (dismissed):

HeikoTheissen:
> But path and `Host` header are not sufficient to determine the protocol (`http:`) of the request URL.

What about this problem?

ralfhandl:
> > But path and `Host` header are not sufficient to determine the protocol (`http:`) of the request URL.
> 
> What about this problem?

The scheme is not necessary inside the batch request because the HTTP message to be routed over the network is the whole batch request, and its scheme has already been used to establish a secured or unsecured connection.

HeikoTheissen:
We need an absolute request URL which must serve as the base URI of a relative `Location` URL. And if the scheme is unclear, we cannot construct one in Example 102.

HeikoTheissen:
> Don't we have that by taking the batch request's scheme, the `Host` header value, and the absolute path from the request line?

If we take the scheme from request, why not the host as well? Why do we need the `Host` header then?

ralfhandl:
> Why do we need the `Host` header then?

Because that's HTTP/1.1 message format, and it is valid to just wrap those into a multipart body.

We can of course debate whether the Host header has any meaning here, or ask how existing implementations use this second form.

ralfhandl (approved):

PR #1918, opened 2024-04-04
Alternative search syntax

HeikoTheissen:


HeikoTheissen (commented):
Instruct clients to evaluate `SearchRestrictions/UnsupportedExpressions` and be prepared for NOT being taken as a search term.

odata-url-conventions/5 Query Options.md
```diff
@@ -2379,6 +2379,14 @@ http://host/service/Products?$search=blue OR green
 ```
 :::
 
+Services MAY implement search based on a different syntax provided they advertise
+this with the annotation `SearchRestrictions/SearchSyntax` defined in
+[OData-VocCap](#ODataVocCap). Clients are advised to evalute this annotation,
+because it governs what result they can expect for a given search request.
+Especially if `SearchRestrictions/Searchable` is false, the service may treat
+keywords that appear in the [OData-ABNF](#ODataABNF) `search` syntax
+as terms to be matched.
```
HeikoTheissen, line 2388: Open issue from TC 2024-04-17:
If `SearchRestrictions/Searchable` is true, may the service treat keywords that it declares as `SearchRestrictions/UnsupportedExpressions` as terms to be matched?
@mikepizzo said "no", @HeikoTheissen said "yes".

ralfhandl (approved):

PR #1926, opened 2024-04-24
Explain unavailability of properties

HeikoTheissen:


odata-protocol/11 Data Service Requests.md
```diff
@@ -163,11 +163,14 @@ Clients MUST be prepared to receive additional properties in an entity
...
 
-Properties that are not available, for example due to permissions, are
-not returned. In this case, the
+Properties that are not available are not returned. If their unavailability
+is due to permissions, the
 [`Core.Permissions`](https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.md#Permissions)
 annotation, defined in [OData-VocCore](#ODataVocCore) MUST be returned
 for the property with a value of `None`.
+If the [`omit-values`](#Preferenceomitvalues) preference is
+applied, an explanatory annotation MUST be returned for every unavailable
```
ralfhandl, line 172: Which annotation should be returned? Do we need a new one, or can this be any annotation?

HeikoTheissen: There can be different annotations, depending on the reason for the unavailability. Most of them are not invented yet.

ralfhandl:
```suggestion
applied, a reason-specific explanatory annotation MUST be returned for every unavailable
```
Something along this line to indicate that it is not a single annotation that we forgot to mention.

ralfhandl (approved):

PR #1927, opened 2024-04-25
Transient entities as parameters

HeikoTheissen:


odata-url-conventions/5 Query Options.md
```diff
@@ -1676,6 +1676,17 @@ $filter=[FirstName,LastName] in [["John","Doe"],["Jane","Smith"]]
 ```
 :::
 
+Transient entities are represented as complex literals with their entity set
```
ralfhandl, line 1679: We should adjust the section title and use "Structured" instead of "Complex".

ralfhandl (approved):

odata-url-conventions/5 Query Options.md
ralfhandl, line 1679: Entity set / context URL is optional. Could also specify the type in case of inheritance.

ralfhandl: Can also use structured literal that represents an entity reference.

mikepizzo: For non-transient, could presumably do any of the following:
1.    {"@id":"Products(1)}                                   // this is actually how I would have expected it to be represented
2.    {"@context":"#Products","ProductId":1}    // alternate representation specifying entity set and key
3.    $root/Products(1)                                      // path syntax

odata-json-format/16 Bound Function.md
```diff
@@ -234,6 +234,25 @@ properties, or just the [entity reference](#EntityReference), as
...
+
+::: example
+Example ##ex: create a quote for a product that does not yet exist
+```json
+POST http://host/service/CreateQuote
+Content-Type: application/json
+
+{
+  "Product": {
+    "@context": "#Products",
```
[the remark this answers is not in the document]
ralfhandl: Special case of passing existing entities: is an object with the key properties "equivalent" to an entity reference?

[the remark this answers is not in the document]
ralfhandl: Maybe add example for passing an entity reference as this is deemed allowed by the majority.

odata-url-conventions/5 Query Options.md
```diff
@@ -1677,6 +1677,16 @@ $filter=[FirstName,LastName] in [["John","Doe"],["Jane","Smith"]]
 ```
 :::
 
+Entities are represented as structured literals as described in [OData-JSON, section 6](#ODataJSON).
+Non-transient entities can alternatively be represented through their [resource path](#ResourcePath).
+
+::: example
+Example ##ex: determine the price of an adhoc-defined product
+```
+http://host/service/Price(Product=@p)?@p={"@context":"#Products","Color":"red"}
```
ralfhandl, line 1686:
```suggestion
http://host/service/Price(Product=@p)?@p={"Color":"red"}
```

odata-json-format/16 Bound Function.md
```diff
@@ -234,6 +234,56 @@ properties, or just the [entity reference](#EntityReference), as
...
+if necessary, its context:
+```json
+POST http://host/service/CreateQuote
+Content-Type: application/json
+
+{
+  "Product": {
+    "@context": "#Products",
+    "ProductID": 14
+  },
```
HeikoTheissen, line 281: @Hubert-Heijkers: Does the reference to section 15.2 justify using the primary key here?

mikepizzo (approved):

PR #1929, opened 2024-05-08
Fixes #398 - Describe creation semantics for singleton media entity.

### LINKED ISSUE #398: Clarify creation of a singleton media entity
11.4.7.1 Create a Media Entity in [Protocol] describes how to create a media entity within a collection. It doesn't mention how to create a singleton contained media entity.

11.4.7.2 Update a Media Entity Stream says that, to update, you would submit a request to the media edit link (the editlink of the entity suffixed with /$value).

Presumably, creation of a singleton contained media entity would be done by calling PUT to the media entity link, which should return 201 created and a location header, and may set other properties of the entity, but this is not called out.

11.4.4 Upsert an Entity calls out that it's not possible to upsert a media entity, which is correct but misleading.  PUT to the media edit link replaces the media. As above, if the entity doesn't already exist it will be created with the new stream value. In either case, additional properties of the entity may be set.

### Proposal

Clarify, in section 11.4.7.1, that a singleton contained media entity is created by calling PUT to the media entity link. If the singleton was null, this should return 201 created and a location header, and may set other properties of the entity.

In 11.4.4 Upsert an Entity, clarify that, while it's not possible to upsert a media entity through a PATCH to the entity url, a PUT to the media edit link replaces the media and, if the entity doesn't already exist, results in the creation of the entity with the new stream value. In either case, additional properties of the media entity may be set when setting the stream value.

Imported from [ODATA-1638](https://issues.oasis-open.org/browse/ODATA-1638)

mikepizzo:
Fixes #398

Clarifies that PUT to the media edit link of a media entity behaves as an upsert.
Clarifies create of a singleton media entity.


HeikoTheissen (dismissed):

odata-protocol/11.4 Data Modification.md
```diff
@@ -550,12 +550,17 @@ Upserts to single-valued navigation properties are possible for
 - payloads including a context URL specifying the entity set or
   contained collection of entities in which the new entity is to be created.
 
-Upserts are not supported against [media
-entities](#RequestingtheMediaStreamofaMediaEntityusingvalue)
-or entities whose keys values are
+Upserts are not supported against entities whose keys values are
```
ralfhandl, line 553: "keys' values" or "key values"?

ralfhandl (approved):

PR #1930, opened 2024-05-08
Fixes #369 -- Describe semantics for changing entity type on update

### LINKED ISSUE #369: Should be able to change type when PATCHing an entity
In [Protocol](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) we say that Entity id and entity type cannot be changed when updating an entity.

[ODATA-1472](https://issues.oasis-open.org/browse/ODATA-1472) introduces the notion of changing the type of a nested complex type, but stopped short of changing the type of an entity due to concerns that an entities type was part of its identity. 

However, updating an entity type doesn't necessarily change the identity of the entity; for example, an employee can become a manager and back.

An entities id is made up of its container plus its key properties. The allowable types that you could change an entity to is already restricted to the hierarchy of the entityset or navigation property in which the entity instance is contained, and the type of that container defines the key property, so there is no concern that changing the type of an entity will change its container nor its key property.  Therefore, an entity's identity doesn't change when you change its type.

Thus, we should allow the semantics defined in [ODATA-1472](https://issues.oasis-open.org/browse/ODATA-1472) to extend to updating the type of an entity (within the hierarchy of the collection in which it is contained).

### Proposal

Allow the semantics defined in [ODATA-1472](https://issues.oasis-open.org/browse/ODATA-1472) to extend to updating the type of an entity (within the hierarchy of the collection in which it is contained).

Imported from [ODATA-1585](https://issues.oasis-open.org/browse/ODATA-1585)

mikepizzo:


odata-protocol/11.4 Data Modification.md
```diff
@@ -306,6 +306,9 @@ request to a URL that identifies the entity. Services MAY restrict
 updates only to requests addressing the [edit URL](#ReadURLsandEditURLs)
 of the entity.
 
+The body of the request MUST be a valid representation of the
```
HeikoTheissen, line 309:
```suggestion
The body of the request MUST be a valid representation of an entity of the
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -360,10 +369,9 @@ The service MAY return success in this case if the specified value
 matches the value of the property. Clients SHOULD use `PATCH` and
 specify only those properties intended to be changed.
 
-Entity-id and entity type cannot be changed when updating an entity.
+The Entity-id cannot be changed when updating an entity.
```
HeikoTheissen, line 372:
```suggestion
The entity-id cannot be changed when updating an entity.
```

ralfhandl (commented):
Fine with the changes suggested by @HeikoTheissen

odata-protocol/11.4 Data Modification.md
mikepizzo: Fixed.

odata-protocol/11.4 Data Modification.md
mikepizzo: fixed

HeikoTheissen (approved):

PR #1945, opened 2024-06-04
Clarify that `in` returns false with an empty list

### LINKED ISSUE #396: In should support empty literal arrays
[ABNF] defines the inexpr as:

inExpr = RWS "in" RWS ( listExpr / commonExpr )

where listExpr is defined as:

listExpr  = OPEN BWS primitiveLiteral BWS *( COMMA BWS primitiveLiteral BWS ) CLOSE

meaning, the list must have at least one primitive literal.

Client code generating the expression may well write the open paren, loop through the elements of the array writing each into the expression, then write a close paren.  If the array is empty, this will result in an empty listExpr.

Any reason not to allow an empty listExpr and force the client code above to handle this as a special case?

### Proposal

Update listExpr in ABNF:

listExpr  = OPEN BWS [primitiveLiteral BWS *( COMMA BWS primitiveLiteral BWS )] CLOSE

Call out that this evaluates to false.
Be clear that this was added in 4.02.

Imported from [ODATA-1633](https://issues.oasis-open.org/browse/ODATA-1633)

mikepizzo:
Fixes #396

### DIFF
--- a/odata-url-conventions/5 Query Options.md
+++ b/odata-url-conventions/5 Query Options.md
@@ -239,8 +239,8 @@ The `null` value is treated as unknown, so if one operand evaluates to
 The `in` operator returns true if the left operand is a member of the
 right operand. The right operand MUST be either a comma-separated list
 of zero or more primitive values, enclosed in parentheses, or a single expression
-that resolves to a collection. If the right operand is an empty list
-of values, the expression returns false.
+that resolves to a collection. If the right operand is an empty collection
+or list of values, the expression returns false.
 
 ##### ##subsubsubsubsec Logical Operator Examples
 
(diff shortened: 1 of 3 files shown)


odata-url-conventions/5 Query Options.md
```diff
@@ -238,8 +238,9 @@ The `null` value is treated as unknown, so if one operand evaluates to
 
 The `in` operator returns true if the left operand is a member of the
 right operand. The right operand MUST be either a comma-separated list
-of primitive values, enclosed in parentheses, or a single expression
-that resolves to a collection.
+of zero or more primitive values, enclosed in parentheses, or a single expression
+that resolves to a collection. If the right operand is an empty list
+of values, the expression returns false.
```
HeikoTheissen, line 243:
```suggestion
that resolves to a collection. If the right operand is an empty list or collection
of values, the expression returns false.
```
If empty lists deserve mention, so do empty collections?

HeikoTheissen (dismissed):

HeikoTheissen (approved):

ralfhandl (approved):

PR #1947, opened 2024-06-06
Implicit casting

### LINKED ISSUE #405: Casting of common expressions
The section [OData-URL, Common Expression Syntax](https://oasis-tcs.github.io/odata-specs/odata-url-conventions/odata-url-conventions.html#CommonExpressionSyntax) contains no rules for casting of expressions that are part of a bigger expression. (It only contains rules for "numeric promotion".)

For expressions with a given signature (arithmetic operators and functions), it is unclear whether arguments are implicitly casted (so that `2 add '3'` would evaluate to 5) or whether a wrong argument type makes the expression invalid.

For the logical operators `eq`, `ne`, `lt`, `le`, `gt`, `ge` and `in`, it is unclear whether
- arguments of different types make the expression invalid
- or (for `eq`, `ne` and `in`) arguments of different types (numbers, strings, booleans, enum values) are always unequal
- or arguments are casted before comparison, but in which direction? `'4.0'  eq 4` is true when casted to numbers but false when casted to strings.

HeikoTheissen:
Fixes #405


odata-url-conventions/5 Query Options.md
```diff
@@ -125,6 +128,19 @@ and only to itself.
...
+'4.0' eq 4
+```
+evalutes to false, because rule 2 for the [`cast`](#cast) function is used to cast
+both operands to strings, and
+```
+'4.0' eq '4'
+```
+is false.
+:::
+
```
HeikoTheissen, line 143: This is very peculiar. But the `cast` functions allows casting in both directions, so we need a precedence rule here. If casting rule 3 was applied, `4.0 eq 4` would be true.

docs/odata-url-conventions/odata-url-conventions.md
```diff
@@ -1549,7 +1549,10 @@ to the `null` value using the [`eq`](#Equals) and [`ne`](#NotEquals)
 operators.
 
 When applied to operands of numeric types, [numeric
-promotion](#NumericPromotion) rules are applied.
+promotion](#NumericPromotion) rules are applied. Otherwise, if the operands are
+primitive values of different types, the first five rules for the [`cast`](#cast)
+function are applied in the given order until the operands can be casted to a common
```
ralfhandl, line 1554:
```suggestion
function are applied in the given order until the operands can be cast to a common
```

docs/odata-url-conventions/odata-url-conventions.md
```diff
@@ -1698,85 +1714,89 @@ of zero or more primitive values, enclosed in parentheses, or a single expressio
 that resolves to a collection. If the right operand is an empty collection
 or list of values, the expression returns false.
 
+If the left operand and a member of the right operand are primitive values of
+different types, they are implicitly casted to a common type like for the [`eq`](#Equals)
```
ralfhandl, line 1718:
```suggestion
different types, they are implicitly cast to a common type like for the [`eq`](#Equals)
```

odata-url-conventions/5 Query Options.md
```diff
@@ -93,7 +93,10 @@ to the `null` value using the [`eq`](#Equals) and [`ne`](#NotEquals)
 operators.
 
 When applied to operands of numeric types, [numeric
-promotion](#NumericPromotion) rules are applied.
+promotion](#NumericPromotion) rules are applied. Otherwise, if the operands are
+primitive values of different types, the first five rules for the [`cast`](#cast)
+function are applied in the given order until the operands can be casted to a common
```
ralfhandl, line 98:
```suggestion
function are applied in the given order until the operands can be cast to a common
```

odata-url-conventions/5 Query Options.md
ralfhandl, line 1: In section [5.1.1.1 Logical Operators](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part2-url-conventions/odata-v4.02-csd01-part2-url-conventions.html#LogicalOperators) call out that values of different primitive types - after numeric promotion - cannot be compared / are always unequal

ralfhandl: Remove sentence on "cast is optional for first five rules and rule seven"

HeikoTheissen: > Remove sentence on "cast is optional for first five rules and rule seven"

Omitting the `cast` when using a value as a function parameter still makes sense for me.

odata-url-conventions/5 Query Options.md
```diff
@@ -93,7 +93,8 @@ to the `null` value using the [`eq`](#Equals) and [`ne`](#NotEquals)
 operators.
 
 When applied to operands of numeric types, [numeric
-promotion](#NumericPromotion) rules are applied.
+promotion](#NumericPromotion) rules are applied. Otherwise primitive
+operands of different types are always unequal.
```
ralfhandl, line 97: This section is for all logical operators, not just equality. Don't know how to express this, maybe we have to add the statement about different types to every section below.

HeikoTheissen: The new sentence comes _before_ the `Equals` section, therefore it applies to the other logical operators up to 5.1.1.1.6 as well.

ralfhandl: Then let's be explicit, because "unequal due to different types" does not help in evaluating "greater than" comparisons. We should explicitly state that this evaluates to false if the types are different and not both numeric.

HeikoTheissen: I suggest:
* `'foo' eq 4` and `'foo' in (4)` are false
* `'foo' ne 4` is true
* `'foo' lt 4`, `'foo' le 4`, `'foo' gt 4`, `'foo' ge 4` are invalid

ralfhandl: "invalid" means "false"?

docs/odata-url-conventions/odata-url-conventions.md
```diff
@@ -1555,6 +1555,9 @@ The `eq`, `ne`, and `in` operators can be used with collection-valued
 operands, and the `eq` and `ne` operators can be used with operands of a
 structured type.
 
+The Boolean operators `and`, `or` and `not` MUST have Boolean operands,
+otherwise the expression is invalid.
```
ralfhandl, line 1559: How are services supposed to treat invalid expressions?
- Treat as "not true" and continue, or
- Fail the request?

ralfhandl (changes requested):

odata-url-conventions/5 Query Options.md
HeikoTheissen: If different instances of `foo` have different types, the property must be `Edm.PrimitiveType` or `Edm.Untyped`, right?

docs/odata-url-conventions/odata-url-conventions.md
HeikoTheissen: See #1951

odata-url-conventions/5 Query Options.md
```diff
@@ -882,13 +893,15 @@ Edm.Boolean matchespattern(Edm.String,Edm.String,Edm.String)
 ```
 
 The second parameter MUST evaluate to a string containing an
-[ECMAScript](#_ECMAScript) (JavaScript) regular expression. The 
-`matchespattern` function returns true if the first parameter evaluates
+[ECMAScript](#_ECMAScript) (JavaScript) regular expression, otherwise the function
+returns `null`. The
```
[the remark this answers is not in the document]
HeikoTheissen: Distinguish between `false` = absence of a pattern and `null` = no statement about any pattern

odata-url-conventions/5 Query Options.md
```diff
@@ -882,13 +893,15 @@ Edm.Boolean matchespattern(Edm.String,Edm.String,Edm.String)
...
+returns `null`. The 
+`matchespattern` function returns `true` if the first parameter evaluates
 to a string matching that regular expression, using syntax and semantics
 of ECMAScript regular expressions, otherwise it
-returns false.
+returns `false`.
 If the optional third parameter is provided, it MUST evaluate to a string
-consisting of ECMAScript regular expression flags to modify the match.
+consisting of ECMAScript regular expression flags to modify the match, otherwise
+the function returns `null`.
```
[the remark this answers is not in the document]
HeikoTheissen: See previous comment.

odata-url-conventions/5 Query Options.md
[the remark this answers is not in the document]
ralfhandl: In which other places do we do that?

As far as I can see we typically state when the function returns `true` and otherwise it returns `false`.

ralfhandl (dismissed):

odata-url-conventions/5 Query Options.md
```diff
@@ -192,55 +198,55 @@ with language-dependent order with the term
...
 value whose flag(s) are set on the left operand.
 
 The `null` value is treated as unknown, so if one operand evaluates to
 `null`, the `has` operator returns `null`.
 
 ##### ##subsubsubsubsec In
 
-The `in` operator returns true if the left operand is a member of the
-right operand. The right operand MUST be either a comma-separated list
+The `in` operator returns `true` if the left operand [equals](#Equals) a member of
```
ralfhandl, line 245: Rephrase "equals must be true" or similar

odata-url-conventions/5 Query Options.md
ralfhandl, line 1: Add text that allows services to fail early if they can detect that the expression is "statically" invalid according to metadata.

ralfhandl: Nice text and example!

odata-url-conventions/5 Query Options.md
```diff
@@ -816,7 +845,7 @@ hassubset([4,1,3,1],[1,1])
 :::
 
 ::: example
-Example ##ex: `hassubset` expression that returns false: 1 appears only
+Example ##ex: `hassubset` expression that returns `false`: 1 appears only
```
ralfhandl, line 848:
```suggestion
Example ##ex: `hassubset` expression that returns `false`: `1` appears only
```

ralfhandl (commented):
I like it! 

Only one minor nit.

ralfhandl (dismissed):

ralfhandl:
If services _can_ detect a discrepancy, they _MUST_ raise an error.

odata-url-conventions/5 Query Options.md
```diff
@@ -90,7 +90,7 @@ http://host/service/People?$filter=Age gt '50'
 ```
 the expression would always evaluate to `null` because the age 50 is erroneously
 given as a string and the [`$filter`](#SystemQueryOptionfilter) would return an empty result, although this is
-really the result of a typing error. That's why a "type mismatch" error SHOULD
+really the result of a typing error. That's why a "type mismatch" error must
```
ralfhandl, line 93:
```suggestion
really the result of a typing error. That's why a "type mismatch" error MUST
```
If we really mean it, we MUST use uppercase according to https://www.rfc-editor.org/rfc/rfc8174.html#section-1

HeikoTheissen: But not in an example (which is non-normative).

ralfhandl (approved):

ralfhandl:
Needs a rebuild

ralfhandl (approved):

PR #1948, opened 2024-06-06
Publish OData Vocabularies Version 4.0

HeikoTheissen:
- [x] Rebuild after merging #1949


docs/odata-vocabularies/odata-vocabularies.md
ralfhandl, line 1: Can we link to the XML files within Github, that is

    https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.xml#L69

instead of

   https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Core.V1.xml#L69:~:text=%3CTerm%20Name=%22-,ODataVersions,-%22

which does not position on that line or text in Chrome on my machine and does not work at all in Safari.

ralfhandl (changes requested):

docs/odata-vocabularies/odata-vocabularies.md
HeikoTheissen: You mean revert the change we made with https://github.com/oasis-tcs/odata-vocabularies/pull/139? (It worked back then.)

ralfhandl: No, I mean use the path prefix 

    https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies

instead of the one for `github.io`.

ralfhandl (approved):

PR #1952, opened 2024-06-18
Describe returning a 200 response from a DELETE request

### LINKED ISSUE #380: Relax constraint that DELETE Cannot return data
The current wording in [11.4.5 Delete an Entity](http://docs.oasis-open.org/odata/odata/v4.01/os/part1-protocol/odata-v4.01-os-part1-protocol.html#sec_DeleteanEntity) in [Protocol] states:

On successful completion of the delete, the response MUST be 204 No Content and contain an empty body.

This is overly restrictive.  HTTP allows Delete to return a response, and since we added a preference header for clients to request a response from requests that typically may not return responses, we should allow clients to request, and services to return, responses from DELETE (i.e., if return=representation preference is specified).

The response should be a representation of the deleted entity.

### Proposal

Allow clients to request, and services to return, responses from DELETE (i.e., if return=representation preference is specified).

The response should be a representation of the deleted entity (i.e., a tombstone) 

If the service supports $select/$expand on the DELETE, standard semantics apply (the same as would be specified in a delta response or delta update payload.)

Also clarify, in the collection case, the response minimally includes the items requested to be deleted that were NOT deleted, and MAY also include tombstones for the ones that were successfully deleted.

Imported from [ODATA-1436](https://issues.oasis-open.org/browse/ODATA-1436)

mikepizzo:
Fixes #380


odata-protocol/11.4 Data Modification.md
```diff
@@ -602,8 +602,10 @@ annotating the singleton with the term `Capabilities.DeleteRestrictions`
 (nested property `Deletable` with value `true`) defined in
 [OData-VocCap](#ODataVocCap).
 
-On successful completion of the delete, the response MUST be
-[`204 No Content`](#ResponseCode204NoContent) and contain an empty body.
+On successful completion of the delete, the response MUST either be
+[`204 No Content`](#ResponseCode204NoContent) and contain an empty body,
+or [`200 OK`](#ResponseCode200OK) and contain a representation of a
+`deleted entity` according to the specified format.
```
HeikoTheissen, line 608:
```suggestion
deleted entity according to the specified format.
```
Or why was that in monospace?

ralfhandl: I checked in Protocol and JSON Format and we consistently use standard font for "deleted entity".

ralfhandl (approved):

HeikoTheissen (approved):

PR #1953, opened 2024-06-18
Add clarifications around untyped collections

### LINKED ISSUE #367: Further clarify restrictions around open types
[ODATA-1498](https://issues.oasis-open.org/browse/ODATA-1498) clarifies that "A collection of primitive values that occurs in a property of type Edm.Untyped is interpreted as a collection of Edm.Boolean, Edm.String, and Edm.Decimal, depending on the JavaScript type."

This rule is true because we can't specify the type of a primitive value within a collection. However, the wording is misleading, because:
1) It's a bit ambiguous what "occurs in a property of type Edm.Untyped" references – is that to say, that the property *is* an untyped collection of primitives, or is an untyped collection that contains a collection of untyped primitives, or...? 

2) It doesn't mention Collection(Edm.Untyped)

3) It implies that the *collection* is a collection of one of the javascript types, as opposed to an individual value within any untyped collection being interpreted as one of these types.  i.e., the collection itself is untyped, and can contain primitives of any of these three types.

Better wording would be something like "A primitive value within an untyped collection is interpreted as a Edm.Boolean, Edm.String, or Edm.Decimal."

In addition, we should note that a collection nested within a collection will always be an untyped collection; this is because there is no place to put a type annotation for the collection, and we say that, once you enter the untyped world, you remain untyped unless type information is explicitly supplied.

 

### Proposal

1) Close [ODATA-1498](https://issues.oasis-open.org/browse/ODATA-1498) with no further action. It clarifies what was originally intended, and that clarification has been applied. Use this issue to track restrictions around how primitives are interpreted within an untyped collection.

2) Add a note to [OData-JSON] that "A primitive value within an untyped collection is interpreted as a Edm.Boolean, Edm.String, or Edm.Double."

3) Clarify that an enum value within an untyped collection should be represented as a string, using the member name.

4) Clarify that any collection within an untyped collection is itself untyped.

Imported from [ODATA-1580](https://issues.oasis-open.org/browse/ODATA-1580)

mikepizzo:
Fixes #367


odata-json-format/7 Structural Property.md
```diff
@@ -183,7 +183,8 @@ collections.
 
 The value of a property of type `Collection(Edm.Untyped)`MUST
 be a collection, and it MAY contain any combination of primitive values,
-structural values, and collections.
+structural values, and collections. Enum values within an untyped 
+collection SHOULD be represented as a string, using the member name.
```
HeikoTheissen, line 187:
```suggestion
structural values, and collections. Enumeration values within an untyped 
collection SHOULD be represented as a string, using the `enumerationMember`.
```
Use the same wording as in section 7.1.

odata-json-format/7 Structural Property.md
```diff
@@ -195,6 +196,8 @@ they are annotated with the
 control information, in which case they MUST conform to the type
 described by the control information.
 
+Collections within an untyped collection are themselves untyped
```
HeikoTheissen, line 199:
```suggestion
Collections directly contained within an untyped collection are themselves untyped 
```

odata-json-format/7 Structural Property.md
```diff
@@ -138,8 +138,8 @@ Example ##ex: partial collection of strings with next link
 ```
 :::
 
-A collection of primitive values that occurs in a property of type `Edm.Untyped`
-is interpreted as a collection of `Edm.Boolean`, `Edm.String`, and `Edm.Decimal` values,
+A primitive value within an untyped collection of primitive values
```
ralfhandl, line 141:
```suggestion
A primitive value within an untyped collection
```
Can untyped collections contain a mix of primitive, object, and array values?

ralfhandl: Why have this here?

Better move it to section "Untyped Value" and link to this section here.

odata-json-format/7 Structural Property.md
ralfhandl: In other specs we only use the long form "enumeration value".

odata-json-format/7 Structural Property.md
ralfhandl: Isn't the collection value of an object property in an untyped collection also untyped?

If so, the "directly" would mislead me.

HeikoTheissen: Can't I specify `@type` control information even in an untyped collection?

ralfhandl: 🤷🏻‍♂️

mikepizzo: a structured item within the untyped collection could have a collection-valued property whose @type was specified. so "directly contained" is more precise.

odata-json-format/7 Structural Property.md
mikepizzo: Moved.

odata-json-format/7 Structural Property.md
mikepizzo: Renamed to "directly contained"

odata-json-format/7 Structural Property.md
mikepizzo: fixed.

ralfhandl (approved):

PR #1958, opened 2024-06-20
Describe semantics of etags within a delta payload.

### LINKED ISSUE #358: Clarify semantics of etag in delta requests/responses
We define that etags may be returned in a response for a single entity or as control information for each entity in a response containing a collection of entities.

These etags can be used in an if-match header in order to enforce optimistic concurrency when updating an instance of one of these entities.

We don't, however, describe how this applies to delta response or update payloads.

### Proposal

Services that support optimistic concurrency (i.e., that would return etag control information in a GET to a collection of entities) should include the current etag value in a delta response for added/updated entities (they could include for deleted entities as well, which would tell the receiver the last version of the entity when it was deleted, but it's of dubious value).

Clients can specify an etag value in a delta update, in which case the update should only apply if the etag matches the current etag value in the server. If it does not match, then the service should either fail the entire operation, or not apply the change and mark it as a datamodificationerror in the response, depending on the continueonerror behavior.

The special value `"@odata.etag":"*"` can be used to request overriding optimistic concurrency control, identical to an `If-Match:*` header in a PATCH request to a single entity.

If an etag value is present, this is interpreted as an update of an existing entity, and fails if the entity doesn't exist yet.

Without an etag value it can be interpreted as an upsert if the service doesn't require optimistic concurrency for that entity set.

Request body example:
```json
{
  "@context" : "#$delta",
  "value" : [
    {
      "Id" : "New Entity",
      "Name" : "Foo",
      "BudgetCurrency" : "USD",
      "Budget" : 555.55
    },
    {
      "@odata.etag" : "W/\"197707241234132000000.0000000 \"",
      "Id" : "Existing Entity 1",
      "Name" : "Bar"
    },
    {
      "@odata.etag" : "W/\"197707240839125000000.0000000 \"",
      "@odata.removed" : {},
      "Id" : "Existing Entity 2",
    }
  ]
}

```

Imported from [ODATA-1549](https://issues.oasis-open.org/browse/ODATA-1549)

mikepizzo:
Fixes #358


odata-protocol/11.4 Data Modification.md
```diff
@@ -1040,6 +1040,10 @@ through the `ContentIDSupported` property of the
 [`Capabilities.DeepUpdateSupport`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#DeepUpdateSupportType)
 term, both defined in [OData-VocCap](#ODataVocCap).
 
+Clients MAY specify [ETag](#UseofETagsforAvoidingUpdateConflicts) values
+obtained from previous requests when updating entities within a collection. If an ETag 
+is provided that does not match the ETag value of the entity being updated, or if the entity does not already exist, services that support ETags MUST NOT apply the change and SHOULD indicate a response code of `412 Precondition Failed`. The special value `\*` can be used to match any existing entity but fail if the entity does not already exist.
```
HeikoTheissen, line 1045:
```suggestion
is provided that does not match the ETag value of the entity being updated, or if the entity does not already exist, services that support ETags MUST NOT apply the change and SHOULD indicate a response code of [`412 Precondition Failed`](#ResponseCode412PreconditionFailed). The special value `*` can be used to match any existing entity but fail if the entity does not already exist.
```
Or does the special value really contain a backslash?

HeikoTheissen (changes requested):

odata-protocol/11.4 Data Modification.md
ralfhandl: No, the backslash was an artifact when migrating from Jira issue [ODATA-1549](https://issues.oasis-open.org/browse/ODATA-1549) which does not have this backslash. Fixed the Github issue description.

ralfhandl: This text only describes the `continue-on-error=false` case, without calling that out.

It also doesn't state that the ETags are per entity, using the `@odata.etag` control information.

ralfhandl (changes requested):

odata-protocol/11.4 Data Modification.md
mikepizzo: Good catch on the backslash -- I actually thought it *was* part of the string due to some requirements on ETag format, but happy to see it is not intended.

I struggled a lot with the continue-on-error stuff. We don't introduce continue-on-error until later in the section, so my thought was that we would provide the general context (that the error is 412 Precondition Failed, without saying *how* it was returned, or whether it stops processing) and then, when the reader reads the section on continue-on-error they would just apply that knowledge to understand how the error case was handled for continue-on-error true or false.

HeikoTheissen: Suggest to add an item to the bulleted list that follows:
* Unmatched ETags within the request MUST be represented in the response as as either entities or entity references, annotated with the term `Core.DataModificationException` with a `responseCode` value of 412.

And augment the last sentence:

If the `continue-on-error` preference has not been specified, and the service is unable to apply all of the changes in the request, then it MUST return an error response _with status code 412, if appropriate,_ and MUST NOT apply any of the changes specified in the request payload.

ralfhandl: That would nicely capture it!

mikepizzo: See if the current wording helps -- I broke out the section on error handling and reference it from the etag section from proper error handling.

odata-protocol/11.4 Data Modification.md
```diff
@@ -1078,18 +1086,29 @@ If an individual change fails due to a failed dependency, it MUST be
...
+are deleted. 
+
+For each entity being updated, clients MAY specify an [ETag](#UseofETagsforAvoidingUpdateConflicts)
+value obtained from a previous request. If an ETag is provided that does not match the ETag 
+value of the entity being updated, or if an ETag is provided for an entity that does not 
+currently exist, then services that support ETags MUST NOT apply the 
+change and instead report a `412 Precondition Failed` error according to the appropriate [`continue-on-error`](#ErrorHandlingwhenReplacingaCollectionofEntities) behavior. 
+The special ETag value `*` can be used to match any existing entity but fail if the entity does not already exist.
+
+#### ##subsubsubsec Error Handling when Replacing a Collection of Entities
```
HeikoTheissen, line 1104: This section lacks the changes I suggested with https://github.com/oasis-tcs/odata-specs/pull/1958#discussion_r1648484613.

ralfhandl: Added it to section "Update a Collection of Entities".

odata-protocol/11.4 Data Modification.md
```diff
@@ -1078,18 +1088,31 @@ If an individual change fails due to a failed dependency, it MUST be
 annotated with the term [`Core.DataModificationException`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#DataModificationException) and SHOULD specify
 a `responseCode` of `424` ([Failed Dependency](#ResponseCode424FailedDependency)).
 
-Alternatively, the verb `PUT` can be used, in which case the request
-body MUST be the representation of a collection of entities. In this
+### ##subsubsec Replace a Collection of Entities
```
ralfhandl, line 1091: Introducing a new subsection next to "Update a Collection of Entities" requires to repeat or reference everything from the preceding section that also applies here.

odata-protocol/11.4 Data Modification.md
HeikoTheissen: Status code 412 must also be mentioned for the case where the `continue-on-error` preference has not been applied.

odata-protocol/11.4 Data Modification.md
```diff
@@ -1078,18 +1088,31 @@ If an individual change fails due to a failed dependency, it MUST be
...
+change and instead report a `412 Precondition Failed` error according to the appropriate [`continue-on-error`](#ErrorHandlingwhenReplacingaCollectionofEntities) behavior. 
+The special ETag value `*` can be used to match any existing entity but fail if the entity does not already exist.
+
+#### ##subsubsubsec Error Handling when Replacing a Collection of Entities
+
+If the `continue-on-error` preference has been applied and any errors occur 
+in processing the changes, then a response MUST be returned regardless of the
 [`return`](#Preferencereturnrepresentationandreturnminimal) preference, and MUST
 contain the full membership and values of the collection as it exists in
 the service.
```
ralfhandl, line 1113: How are skipped data modifications expressed in this full membership response?

We either have to add adapted copies of the relevant parts of the sibling subsubsection for "... Updating ..." or extract the common parts into separate subsub(sub?)sections.

ralfhandl (changes requested):

odata-protocol/11.4 Data Modification.md
```diff
@@ -1058,9 +1067,10 @@ of `deleted`, the value of `failedOperation` MUST be `delete`, otherwise
 be represented in the response as deleted entities annotated with the term
 `Core.DataModificationException` with a `failedOperation` value of
 `insert`.
-- Failed updates within the request SHOULD
+- Failed updates within the request MUST
```
ralfhandl, line 1070: Why did we have "SHOULD" for updates?

I don't think we can leave that at SHOULD and allow services to silently ignore updates, especially if they "only" fail due to an outdated ETag.

odata-protocol/11.4 Data Modification.md
ralfhandl: With "stop on first error" the rules in the referenced section on "Upsert an Entity" apply, which in turn references sections "Create an Entity" and "Update an Entity". These sections already describe the If-Match behavior.

odata-protocol/11.4 Data Modification.md
```diff
@@ -1040,11 +1040,20 @@ through the `ContentIDSupported` property of the
 [`Capabilities.DeepUpdateSupport`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#DeepUpdateSupportType)
 term, both defined in [OData-VocCap](#ODataVocCap).
 
+For each entity being updated or removed, clients MAY specify an [ETag](#UseofETagsforAvoidingUpdateConflicts) value obtained from a previous request. 
+If an ETag is provided that does not match the ETag value of the entity being updated or removed, 
+or if an ETag is provided when adding or updating an entity that does not currently exist, 
+then services that support ETags MUST NOT apply the change and instead report a 
+`412 Precondition Failed` error according to the appropriate [`continue-on-error`](#ErrorHandlingwhenUpdatingaCollectionofEntities) behavior.
```
ralfhandl, line 1047: This paragraph seems out of place and should better move to the "Error Handling ..." section.

odata-protocol/11.4 Data Modification.md
mikepizzo: I suspect we were thinking that the inclusion of the entity with its current values was sufficient -- whether the values had changed because of server logic, an exogenous change, or an error, the client processing should be the same (to apply the current values). If there is an etag then it should be updated, so the client can retry the request.  So, it's not the end of the world if the service doesn't mark the returned entity with a data modification exception.

We can debate whether that logic holds, but I suspect that was our reasoning when we authored the original wording.

odata-protocol/11.4 Data Modification.md
mikepizzo: That was my intention -- to include everything in this section that is relevant to replacing a collection of entities. Turns out, that's not much -- most of the section as written was really about updating the collection using a delta payload, which is why I thought it made sense to break replacing the collection into its own section.  What is relevant to replacing a collection is just:

1) Use PUT
2) Request is the entire collection, things omitted are considered removed
3) You can use ETags
4) If an error occurs, either apply no changes or return a representation showing the current membership.

Is there something else that was relevant to replacing a collection of entities in the original text that is missing from this new section?

odata-protocol/11.4 Data Modification.md
mikepizzo: By showing the current values -- the section on returning a "compensating" delta payload with data modification exceptions was specific to the delta update case.

Agree that this could be fleshed out more, but I think current wording is as complete as previous wording, and less confusing (because the reader doesn't have to guess which parts apply to PATCH versus PUT).

odata-protocol/11.4 Data Modification.md
mikepizzo: By stating "according to the appropriate `continue-on-error` behavior, my intent was to capture both continue-on-error = true (don't make the change to the record and return 412 for the failed record) or continue-on-error = false (don't make any changes and return 412 for the response header, as outlined in the if-not-match section).

That is, the error for not matching is 412 Precondition Failed. Just like any other error that occurs for an individual entity, where the error is returned depends on continue-on-error. If True, then it's returned for the entity (just like any other error that prevented an individual update would be returned), if False then it's returned for the response (just like any other individual error that resulted in the whole operation failing due to continue-on-error=false would be returned as the response code for the whole operation).  There's nothing special about handling a 412, versus any other error that might prevent an individual entity from being updated.

odata-protocol/11 Data Service Requests.md
```diff
@@ -1434,5 +1434,9 @@ A delta payload represents changes to a known state. A delta payload
 includes added entities, changed entities, and deleted entities, as well
 as a representation of added and removed relationships.
 
+Services that support the use of [ETags](#UseofETagsforAvoidingUpdateConflicts) 
+for optimistic concurrency SHOULD return ETag values for added or changed entities
```
ralfhandl, line 1438:
```suggestion
for optimistic concurrency control SHOULD return ETag values for added or changed entities 
```

odata-protocol/11.4 Data Modification.md
ralfhandl: Use of `@etag` on individual entities is surprising. I'd treat this as a new feature.

odata-protocol/11.4 Data Modification.md
```diff
@@ -1078,18 +1088,36 @@ If an individual change fails due to a failed dependency, it MUST be
...
-been specified, and the request returns a success response code, then a
-response MUST be returned regardless of the
+are deleted. 
+
+For each entity being updated, clients MAY specify an [ETag](#UseofETagsforAvoidingUpdateConflicts)
+value obtained from a previous request. If an ETag is provided that does not match the ETag 
+value of the entity being updated, or if an ETag is provided for an entity that does not 
+currently exist, then services that support ETags MUST NOT apply the 
+change and instead report a `412 Precondition Failed` error according to the appropriate [`continue-on-error`](#ErrorHandlingwhenReplacingaCollectionofEntities) behavior. 
+The special ETag value `*` can be used to match any existing entity but fail if the entity does not already exist.
```
ralfhandl, line 1110: This is a new feature and needs to be called out in the "Changes" section.

odata-protocol/1 Introduction.md
```diff
@@ -38,6 +38,7 @@ Upserts to single-valued non-containment navigation properties|
 [Section ##UpdateaComplexProperty]| 
 Setting a complex property to a different type| 
 [534](https://github.com/oasis-tcs/odata-specs/issues/534)
+[Section ##ReplaceaCollectionofEntities]| Describe semantics for using `continue-on-error` when replacing a collection of entities. | [358](https://github.com/oasis-tcs/odata-specs/issues/358)
```
ralfhandl, line 41:
```suggestion
[Section ##ReplaceaCollectionofEntities]| Semantics of `continue-on-error` when replacing a collection of entities | [358](https://github.com/oasis-tcs/odata-specs/issues/358)
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1043,85 +1043,93 @@ term, both defined in [OData-VocCap](#ODataVocCap).
...
 represents failed changes in the delta response as follows:
-- Failed deletes in the request MUST be
-represented in the response as either entities or entity references,
-annotated with the term `Core.DataModificationException`, see
+- Failed deletes in the request MUST be represented in the response as either entities 
+or entity references, annotated with the term `Core.DataModificationException`, see
 [OData-VocCore](#ODataVocCore). If the deleted entity specified a reason
-of `deleted`, the value of `failedOperation` MUST be `delete`, otherwise
-`unlink`.
+of `deleted`, or the target collection is an entity set or contained navigation property,
```
ralfhandl, line 1068:
```suggestion
of `deleted`, or the target collection is an entity set or containment navigation property,
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1043,85 +1043,93 @@ term, both defined in [OData-VocCap](#ODataVocCap).
...
-MUST represented in the response as added links annotated with the term
-`Core.DataModificationException` with a `failedOperation` value of
-`unlink`.
-- Collections within the request MUST be
-represented in the response as a collection with the current values and
-membership of the collection as it exists in the service after
-processing the request.
+- Failed updates within the request SHOULD be annotated in the response with 
+the term `Core.DataModificationException` with a `failedOperation` value of `update`.
+- Failed added links within the request MUST represented in the response as
```
ralfhandl, line 1076:
```suggestion
- Failed added links within the request MUST be represented in the response as 
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1043,85 +1043,93 @@ term, both defined in [OData-VocCap](#ODataVocCap).
...
-- Collections within the request MUST be
-represented in the response as a collection with the current values and
-membership of the collection as it exists in the service after
-processing the request.
+- Failed updates within the request SHOULD be annotated in the response with 
+the term `Core.DataModificationException` with a `failedOperation` value of `update`.
+- Failed added links within the request MUST represented in the response as 
+deleted links annotated with the term `Core.DataModificationException` 
+with a `failedOperation` value of `link`.
+- Failed deleted links within the request MUST represented in the response as
```
ralfhandl, line 1079:
```suggestion
- Failed deleted links within the request MUST represented be in the response as 
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1043,85 +1043,93 @@ term, both defined in [OData-VocCap](#ODataVocCap).
...
 specified in the request payload.
 
+If the `continue-on-error` preference has been applied and any errors occur 
+in processing the changes, then a response MUST be returned regardless of the
+[`return`](#Preferencereturnrepresentationandreturnminimal) preference, and MUST
+contain the full membership and values of the collection as it exists in
+the service, as follows:
+- Entities missing in the request that cannot be removed from the collection 
+MUST be represented in the response as either entities or entity references,
+and MAY be annotated with the term `Core.DataModificationException`, see
```
ralfhandl, line 1122:
```suggestion
and SHOULD be annotated with the term `Core.DataModificationException`, see
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1043,85 +1043,93 @@ term, both defined in [OData-VocCap](#ODataVocCap).
...
+[OData-VocCore](#ODataVocCore). If the target collection is an entity set or 
+containment navigation property, then the value of `failedOperation` MUST be 
+`delete`, otherwise `unlink`.
+- Failed inserts within the request MUST NOT be represented in the response.
+- Failed updates within the request MUST be represented in the response with
+their current values and MAY be annotated with the term `Core.DataModificationException`
```
ralfhandl, line 1128:
```suggestion
their current values and SHOULD be annotated with the term `Core.DataModificationException`
```

ralfhandl (commented):
Nice!

Only some word smithing.

odata-protocol/11.4 Data Modification.md
```diff
@@ -1040,59 +1040,96 @@ through the `ContentIDSupported` property of the
 [`Capabilities.DeepUpdateSupport`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#DeepUpdateSupportType)
 term, both defined in [OData-VocCap](#ODataVocCap).
 
+For each entity being updated or removed, clients MAY specify an [ETag](#UseofETagsforAvoidingUpdateConflicts) value obtained from a previous request.
```
ralfhandl, line 1043: We are a bit vague in Protocol regarding "ETag values specified in the request body" or how exactly an "entity representation in the request body includes an ETag".

Can we explicitly state, for example in section "[Use of ETags for Avoiding Update Conflicts](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#UseofETagsforAvoidingUpdateConflicts)", that 

> The representation of Etags within entity representations is format-specific, see for example JSON Format, section [Control Information: etag (odata.etag)](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#ControlInformationetagodataetag).

ralfhandl (approved):
One remark, which can be addressed in this PR or via a new issue.

odata-protocol/11.4 Data Modification.md
ralfhandl: * Created #1971

ralfhandl (approved):

PR #1959, opened 2024-06-21
Fix puppeteer vulnerability

HeikoTheissen:
Merge only after https://issues.chromium.org/issues/347674894 is fixed.


lib/md2html.js
```diff
@@ -23,7 +23,7 @@ filter.stdio(function ({ t, c }, format, meta) {
     level = c[0];
     return content;
   } else if (t === "HorizontalRule") {
-    var content = [{ t, c }];
+    content = [filter.HorizontalRule()];
```
ralfhandl, line 26: Why not `var` here as in the preceding if-block?

ralfhandl: Then please use `let` in both places

docs/odata-protocol/odata-protocol.pdf
ralfhandl, line 1: Why is this PDF included in the PR?

HeikoTheissen: I tested (with one PDF) whether the new puppeteer version generates correct PDF links.

ralfhandl (approved):

PR #1943, opened 2024-05-31
Expressions in action payloads

HeikoTheissen:
Fixes #341

PR #1960, opened 2024-06-24
`/$query` with JSON body

### LINKED ISSUE #371: /$query shall allow Content-Type: application/json
JSON as content type should also be supported by `/$query`:
```java
POST ~/$query
Content-Type: application/json

{"$filter": "a eq 1",
 "$select": "b"}
```
This is especially useful for complex function parameters, that would then be represented as if they were action parameters:
```java
POST ~/Countries('DE')/GetPostcodes
Content-Type: application/json

{"$filter": "ValidityStart gt 2020-01-01",
 "City": "Heidelberg",
 "Address": {"Street": "Hauptstraße", "Number": "1"}}
```
Note that
- the name of the bound function in the URL is without parentheses (because the parameters are in the body)
- parameters with `@expression` suffix are also allowed (#1943)
- the `/$query` segment in the URL is optional and has been omitted here. (There is no risk of misinterpreting a function invocation URL.)

### Proposal

See description

Note: this would allow providing "literal" values to non-binding function parameters of type `Edm.Stream` introduced by [ODATA-1481](https://issues.oasis-open.org/browse/ODATA-1481) because there's now a natural way to provide the content-type in addition to the stream data:

```json
{
  "Thumbnail@mediaContentType": "image/jpeg",
  "Thumbnail": "...base64url encoded value...",
}
```

Imported from [ODATA-1589](https://issues.oasis-open.org/browse/ODATA-1589)

HeikoTheissen:
Fixes #371

Depends on #1943


ralfhandl (approved):

odata-url-conventions/4 Resource Path.md
```diff
@@ -1061,3 +1062,31 @@ This POST request would result from submitting the HTML form
...
 optional).
 :::
+
+With `Content-Type: application/json` query options and function parameters are
+encoded in a request body that represents a JSON object. Its members include the
+individual query options. The name of a system query option MUST have the `$` prefix.
+The value MUST be
+* a JSON number for `$top` and `$skip`, and
+* a JSON string without percent-encoding for all other query options.
+
```
ralfhandl, line 1072: Please insert same example as for the other two content-types:

```
POST http://host/service/People/$query
Content-Type: application/json

{
  "$filter": "LastName eq 'P&G'",
  "$select": "FirstName,LastName"
}
```

ralfhandl (changes requested):

ralfhandl (approved):

ralfhandl (dismissed):

ralfhandl (dismissed):

odata-url-conventions/4 Resource Path.md
```diff
@@ -1061,3 +1062,45 @@ This POST request would result from submitting the HTML form
...
+}
+```
+:::
+
+Members of the JSON object also include parameters
+if the resource path is a function invocation or function import. In this case
+parameters MUST be represented like parameters in an action invocation [OData-JSON, section 18](#ODataJSON),
+and in the resource path parentheses after the function name MUST be omitted.
+
+::: example
```
ralfhandl, line 1092: Add modified copy of example ##funcexpr with implicit parameter aliases

PR #1924, opened 2024-04-22
Clarify key values in upsert along navigation properties

ralfhandl:
- [x] Fixes #402

PR #1968, opened 2024-07-19
Non-insertable and non-updatable properties

### LINKED ISSUE #356: Allow services to validate read-only properties in update payloads
Section [11.4.3, Updating an Entity](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) states: 

> Key and other properties marked as read-only in metadata (including computed properties), as well as dependent properties that are not tied to key properties of the principal entity, can be omitted from the request. If the request contains a value for one of these properties, the service MUST ignore that value when applying the update. Services MUST return an error if an insert or update contains a new value for a property marked as updatable that cannot currently be changed by the user (i.e., given the state of the object or permissions of the user). The service MAY return success in this case if the specified value matches the value of the property. Clients SHOULD use PATCH and specify only those properties intended to be changed.

We defined the semantics this way because we wanted the ability to take a response payload, make some changes, and PUT it back without having to remove properties that were not updatable by the client.

We chose "ignore" because:
1) We wanted to ignore key properties because we wanted the ability to take a response payload from one service and use it to insert into another service.
2) We wanted to ignore computed properties because the computed value could change based on other properties in the payload or on external information.

However, "read-only" is not a strictly defined term – it refers generically to properties that can't be changed by the user and includes not only key and computed properties, but other read-only properties.

For other such read-only properties, it may be useful, and should be valid, for services to validate that the value of the read-only property matches the stored value and, if not, return an error.

### Proposal

In 4.02, allow services to validate for non-key, non-computed read-only values and return an error if they don't match persisted values. 

Imported from [ODATA-1545](https://issues.oasis-open.org/browse/ODATA-1545)

### LINKED ISSUE #357: Encourage clients to specify the correct value for non-key, non-computed read-only values (or omit)
Section [11.4.3, Updating an Entity](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateanEntity) states: 

> Key and other properties marked as read-only in metadata (including computed properties), as well as dependent properties that are not tied to key properties of the principal entity, can be omitted from the request. If the request contains a value for one of these properties, the service MUST ignore that value when applying the update. Services MUST return an error if an insert or update contains a new value for a property marked as updatable that cannot currently be changed by the user (i.e., given the state of the object or permissions of the user). The service MAY return success in this case if the specified value matches the value of the property. Clients SHOULD use PATCH and specify only those properties intended to be changed.

We defined the semantics this way because we wanted the ability to take a response payload, make some changes, and PUT it back without having to remove properties that were not updatable by the client.

We chose "ignore" because:
1) We wanted to ignore key properties because we wanted the ability to take a response payload from one service and use it to insert into another service.
2) We wanted to ignore computed properties because the computed value could change based on other properties in the payload or on external information.

However, "read-only" is not a strictly defined term – it refers generically to properties that can't be changed by the user and includes not only key and computed properties, but other read-only properties.

For other such read-only properties, it may be useful, and should be valid, for services to validate that the value of the read-only property matches the stored value and, if not, return an error.

### Proposal

Encourage clients to specify the correct value for non-key, non-computed read-only values (or omit).

Imported from [ODATA-1548](https://issues.oasis-open.org/browse/ODATA-1548)

ralfhandl:
- [x] Fixes #356
- [x] Fixes #357
- includes #1924


odata-protocol/11.4 Data Modification.md
```diff
@@ -178,13 +178,23 @@ values beyond those specified in the metadata SHOULD NOT be sent in the
...
 
-Properties computed by the service (annotated with the term
-[`Core.Computed`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Computed),
-see [OData-VocCore](#ODataVocCore)) and properties that are tied to
-properties of the principal entity by a referential constraint, can be
-omitted and MUST be ignored if included in the request.
-
-Properties with a defined default value, nullable properties, and
+Non-insertable properties SHOULD be omitted from the request body.
+If they are provided, services MUST either ignore the value in the request body or fail the request if the provided values do not match the service-determined values.
```
ralfhandl, line 182: Non-breaking alternative:
```suggestion
If they are provided, services MUST ignore the value in the request body.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -178,13 +178,24 @@ values beyond those specified in the metadata SHOULD NOT be sent in the
...
 values specified in the request.
 
-Properties computed by the service (annotated with the term
-[`Core.Computed`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Computed),
-see [OData-VocCore](#ODataVocCore)) and properties that are tied to
-properties of the principal entity by a referential constraint, can be
-omitted and MUST be ignored if included in the request.
+Non-insertable properties SHOULD be omitted from the request body.
+If they are provided, services MUST either ignore the values in the request body or fail the request,
+for example if the provided values do not match the service-determined values.
```
ralfhandl, line 183:
```suggestion
If they are provided, services MUST either ignore the values in the request body or fail the request
if the provided values do not match the service-determined values.
```

Options:
1. check metadata and don't look at the value if it is "read-only"
2. check metadata, look at the value, compute default, compare, then reject if different

odata-protocol/11.4 Data Modification.md
```diff
@@ -178,13 +178,24 @@ values beyond those specified in the metadata SHOULD NOT be sent in the
...
+- dependent properties that are tied to non-key properties of the principal entity through a referential constraint [OData-CSDL, section 8.5](#ODataCSDL) (informally: "denormalized" properties),
+- properties annotated with the term
+  [`Core.Computed`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Computed), see [OData-VocCore](#ODataVocCore),
+- properties listed as `NonInsertableProperties` of term [`Capabilities.InsertRestrictions`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#InsertRestrictions), see [OData-VocCap](#ODataVocCap),
+- properties annotated with term
+  [`Core.Permissions`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Permissions), see [OData-VocCore](#ODataVocCore), where the annotation value does not have the `Write` flag.
+
+Services MUST return an error if the request body contains a value for a
+property that in principle can be inserted and currently cannot be inserted by the
+authenticated user (i.e., given the state of the object or permissions of the user).
```
ralfhandl, line 196:
```suggestion
authenticated user (for example, given the state of the object or permissions of the user).
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -178,13 +178,24 @@ values beyond those specified in the metadata SHOULD NOT be sent in the
...
+
```
ralfhandl, line 197:
```suggestion
Services MUST return an error if the request body contains a value for a property that in principle can be specified on insert but cannot currently be set to the specified value, for example, due to permissions or state of the object.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -360,17 +372,26 @@ Updating a principal property that is tied to a dependent entity through
...
-the service MUST ignore that value when applying the update. Services
-MUST return an error if an insert or update contains a new value for a
-property marked as updatable that cannot currently be changed by the
-user (i.e., given the state of the object or permissions of the user).
-The service MAY return success in this case if the specified value
-matches the value of the property. Clients SHOULD use `PATCH` and
-specify only those properties intended to be changed.
+Non-updatable properties SHOULD be omitted from the request body.
+If they are provided, services MUST either ignore the values in the request body or fail the request,
+for example if the provided values do not match the actual values known by the service.
```
ralfhandl, line 377:
```suggestion
If they are provided, services MUST either ignore the values in the request body or fail the request
if the provided values do not match the actual values known by the service.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -360,17 +372,26 @@ Updating a principal property that is tied to a dependent entity through
...
+- properties annotated with the terms
+  [`Core.Computed`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Computed) or [`Core.Immutable`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Immutable), see [OData-VocCore](#ODataVocCore),
+- properties listed as `NonUpdatableProperties` of term [`Capabilities.UpdateRestrictions`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#UpdateRestrictions), see [OData-VocCap](#ODataVocCap),
+- properties annotated with term
+  [`Core.Permissions`](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md#Permissions), see [OData-VocCore](#ODataVocCore), where the annotation value does not have the `Write` flag.
+
+Services MUST return an error if the request body contains a value for a
+property that in principle can be updated and currently cannot be updated by the
+authenticated user (i.e., given the state of the object or permissions of the user), and
+the specified value differs from the current value of the property.
```
ralfhandl, line 392:
```suggestion
Services MUST return an error if the request body contains a value for a
property that in principle can be specified on update but the request cannot currently be executed respecting the specified value, for example, due to permissions or state of the object.
```

Rationale: don't disappoint the client by silently ignoring provided information.

HeikoTheissen (approved):

HeikoTheissen (approved):

PR #1975, opened 2024-08-04
Single quotes in typewriter mode

HeikoTheissen:
Avoid the mysterious CSS and apply https://stackoverflow.com/a/78829349/16462950

### DIFF
--- a/docs/odata-csdl-json/odata-csdl-json.html
+++ b/docs/odata-csdl-json/odata-csdl-json.html
@@ -89,13 +89,13 @@
   <link rel="stylesheet" href="styles/markdown-styles-v1.7.3b.css" />
   <link rel="stylesheet" href="styles/odata.css" />
   <script type="text/javascript">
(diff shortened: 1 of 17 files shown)


ralfhandl:
Looks consistent, but do we use single quotes in typewriter mode anywhere? If so, I would have expected the corresponding HTML file to change, at least `test.html` should have an example with a single quote in typewriter mode.

What am I missing?

HeikoTheissen:
We use single quotes in typewriter in
https://github.com/oasis-tcs/odata-specs/blob/b0de947bd9911ab0f25519da5bc29da7d2edbfce/odata-data-aggregation-ext/6%20Hierarchical%20Transformations.md?plain=1#L49

The HTML file changes only in the `<head>`, which changes the MathJax Javascript code that does the actual formatting. Testing these effects with puppeteer is too complicated for my taste.

ralfhandl (approved):

PR #1980, opened 2024-08-28
Dark mode for HTML documents

HeikoTheissen:


styles/odata.css
```diff
@@ -201,6 +201,21 @@ hr:first-of-type {
 }
 
 @media screen {
+  body {
+    position: relative;
+  }
+
+  #darkmode {
+    position: absolute;
```
HeikoTheissen, line 209: Would prefer `position: fixed`, but see https://stackoverflow.com/questions/78923866.

ralfhandl (approved):

mikepizzo (approved):

PR #1994, opened 2024-09-12
Change related entity with PUT

### LINKED ISSUE #1973: How to change a reference without changing the referenced entity in a PUT request
The resolution of #352 allows related entities in a PUT/PATCH request to be specified by their key properties alone, with all or some non-key properties omitted (R3 or R4). It is not made clear that PUT behaves differently from PATCH here:
* In a PATCH request, the omitted non-key properties are left unchanged.
* In a PUT request, the omitted non-key properties are set to null or a default value.

If a PUT request just wants to change a reference, without also changing the referenced entity, which representations can it use?
* R1: yes
* R2: partially yes, could be viewed as a sloppy version of R1
* R3, R4, R5: no
* R6: Is a mixture between R2 (partially yes) and R5 (no).

HeikoTheissen:
Fixes #1973

### DIFF
--- a/odata-protocol/11.4 Data Modification.md
+++ b/odata-protocol/11.4 Data Modification.md
@@ -545,7 +545,43 @@ nested delta representation to:
   ]
 }
 ```
+:::
 
+::: example
+Example ##ex: When updating an entity with a 4.01 `PUT` request, the target of a
+non-containment navigation property can be replaced without changing either the
+old or the new targeted entity itself. The targeted entity can then only be
+specified by an entity reference (see [OData-JSON, section "Entity Reference"]($$$OData-JSON$$$#EntityReference)):
+```json
+PUT http://host/service/Employees(7)
+Content-Type: application/json
+
+{
+  "FirstName": "Suzanne",
+  "LastName": "Brown-Johnson",
+  "Manager": {
+    "@context": "$metadata#$ref",
+    "@id": "Employees(6)"
+  }
+}
+```
+
+If the targeted entity in the payload contains some structural properties,
+`PUT` resets all its other structural properties. The following therefore does
+not have the desired effect:
+```json
+PUT http://host/service/Employees(7)
+Content-Type: application/json
+
+{
+  "FirstName": "Suzanne",
+  "LastName": "Brown-Johnson",
+  "Manager": {
+    "@id": "Employees(6)",
+    "EmployeeID": 6
+  }
+}
+```
 :::
 
 Clients MAY associate an id with individual nested entities in the
--- a/docs/odata-protocol/odata-protocol.html
+++ b/docs/odata-protocol/odata-protocol.html
@@ -2471,6 +2471,32 @@ http://host/service/$metadata#Customers/$delta</code></pre>
 <span id="cb98-28"><a href="#cb98-28" aria-hidden="true" tabindex="-1"></a>  <span class="ot">]</span></span>
 <span id="cb98-29"><a href="#cb98-29" aria-hidden="true" tabindex="-1"></a><span class="fu">}</span></span></code></pre></div>
 </div>
+<div class="example">
+<p>Example 80: When updating an entity with a 4.01 <code>PUT</code> request, the target of a non-containment navigation property can be replaced without changing either the old or the new targeted entity itself. The targeted entity can then only be specified by an entity reference (see <a href="https://docs.oasis-open.org/odata/odata-json-format/v4.02/odata-json-format-v4.02.html#EntityReference">OData-JSON, section “Entity Reference”</a>):</p>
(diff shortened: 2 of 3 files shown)


odata-protocol/11.4 Data Modification.md
```diff
@@ -545,7 +545,43 @@ nested delta representation to:
   ]
 }
 ```
+:::
+
+::: example
+Example ##ex: When updating an entity with a 4.01 `PUT` request, the target of a
+non-containment navigation property can be replaced without changing either the
+old or the new targeted entity itself. The targeted entity can then only be
+specified by an entity reference (see [OData-JSON, section "Entity Reference"]($$$OData-JSON$$$#EntityReference)):
```
ralfhandl, line 554: "specified as"?

ralfhandl (dismissed):

odata-protocol/11.4 Data Modification.md
```diff
@@ -545,7 +545,43 @@ nested delta representation to:
...
+Content-Type: application/json
+
+{
+  "FirstName": "Suzanne",
+  "LastName": "Brown-Johnson",
+  "Manager": {
+    "@id": "Employees(6)",
+    "EmployeeID": 6
+  }
+}
```
HeikoTheissen, line 583: Or even stricter?
```suggestion
If the targeted entity in the payload does not have the `@context` of an entity reference,
`PUT` resets all its structural properties not contained in the payload. The following therefore does
not have the desired effect:
```json
PUT http://host/service/Employees(7)
Content-Type: application/json

{
  "FirstName": "Suzanne",
  "LastName": "Brown-Johnson",
  "Manager": {
    "@id": "Employees(6)"
  }
}
```

ralfhandl: That's too strict: an object only containing `@id` essentially is an entity reference, even without the `@context`.

ralfhandl: I'd rather remove the `@id` from the "undesired effect" example

odata-protocol/11.4 Data Modification.md
```diff
@@ -545,7 +545,43 @@ nested delta representation to:
...
+```json
+PUT http://host/service/Employees(7)
+Content-Type: application/json
+
+{
+  "FirstName": "Suzanne",
+  "LastName": "Brown-Johnson",
+  "Manager": {
+    "@id": "Employees(6)",
+    "EmployeeID": 6
```
ralfhandl, line 581:
```suggestion
    "EmployeeID": 6
```

HeikoTheissen: We should clarify both alternatives:
* reference plus some properties
* no reference, just some properties.

odata-protocol/11.4 Data Modification.md
```diff
@@ -567,21 +569,29 @@ Content-Type: application/json
...
-PUT http://host/service/Employees(7)
-Content-Type: application/json
-
 {
   "FirstName": "Suzanne",
   "LastName": "Brown-Johnson",
   "Manager": {
+    "@context": "$metadata#$ref",
     "@id": "Employees(6)",
     "EmployeeID": 6
```
ralfhandl, line 581: That is not a valid entity reference, additional properties are not allowed.

ralfhandl: If you remove the `@context`, it becomes an incomplete entity and a valid example for undesired effects.

ralfhandl (changes requested):

ralfhandl (dismissed):

ralfhandl:
Suggestion from @GEricson: use two-column table, with request on the left and resulting entities in the right column.

ralfhandl (dismissed):

ralfhandl (approved):

ralfhandl:
@HeikoTheissen 
- add headings to requests and responses
- explicitly mention sloppy references without context

ralfhandl (approved):

ralfhandl:
TC 2024-10-09: explicitly state that omitted navigation properties imply no change to the related entities also in PUT requests.

HeikoTheissen:
Change the example to a collection-valued navigation property and point out similarities between contained and non-contained.

odata-protocol/11.4 Data Modification.md
```diff
@@ -449,6 +449,10 @@ references specified in a successful update request represents the full
 set of entities to be related according to that relationship and MUST
 NOT include added links, deleted links, or deleted entities.
 
+If a navigation property is absent from a `PUT` or `PATCH` request payload, the referenced
+or contained entity, or the collection thereof, which the navigation property targets
```
ralfhandl, line 453:
```suggestion
or contained entity, or the collection thereof,
```

ralfhandl (commented):
only minor nits, see comments

ralfhandl (dismissed):

ralfhandl (approved):

PR #1997, opened 2024-09-17
Resolve cross-spec references

HeikoTheissen:


meta.yaml
```diff
@@ -5,6 +5,7 @@ OData-Data-Agg: https://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.
 OData-Temporal: https://docs.oasis-open.org/odata/odata-temporal-ext/v4.0/odata-temporal-ext-v4.0.html
 OData-Protocol: https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part1-protocol.html
 OData-URL: https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part2-url-conventions.html
```
HeikoTheissen, line 7: The `OData-URL` property will eventually be replaced by `odata-url-conventions`.

ralfhandl: Why? OData-URL is shorter and expressive enough.

HeikoTheissen: The new syntax uses the directory name to identify another spec document.

ralfhandl (approved):
Cool!

odata-url-conventions/5 Query Options.md
```diff
@@ -1706,7 +1706,7 @@ $filter=[FirstName,LastName] in [["John","Doe"],["Jane","Smith"]]
 ```
 :::
 
-Entities are represented as structured literals as described in [OData-JSON #odata-json-format#Entity].
+Entities are represented as structured literals as described in [OData-JSON, section #odata-json-format#Entity].
```
ralfhandl, line 1709: Why do we have to write "section" everywhere?

HeikoTheissen: We write "section" for internal references as well. [OData-JSON, 6] could also mean a page. But let the TC decide.

ralfhandl: I thought `[subsection ##SubsectionHeading]` and `[example ##aggregation]` were hints for the reference resolver to distinguish between section references and example references, not actual output text.

If I write `[OData-CSDL #odata-json-format#Entity]`, would the text say "OData-CSDL" and the link point to OData-JSON?

Being lazy I'd prefer to write `[#odata-json-format#Entity]` and get both the text "OData JSON, section 6" and the correct link. I'd even be willing to write more hashes if I don't have to repeat the spec name, for example `[###odata-json-format#Entity]`.

HeikoTheissen: > If I write `[OData-CSDL #odata-json-format#Entity]`, would the text say "OData-CSDL" and the link point to OData-JSON?

I use this "redundancy" to write `[OData-CSDL, #odata-csdl-json#Annotations]`, because the link goes to the JSON variant, not the XML variant.

ralfhandl: I want to see it in the output text, I'd prefer not having to write it _unless_ we want to allow referencing _examples_ in other documents in the future.

HeikoTheissen: Long syntax `[OData-Protocol, section #OData-Protocol#SectionHeading]`. (Can also be used for examples.)
Short variant for section references: `[#OData-Protocol#SectionHeading]`

ralfhandl (approved):
Excellent!

ralfhandl (approved):

ralfhandl (approved):

ralfhandl (approved):

PR #2008, opened 2024-09-27
No duplicate nested query options in $expand

### LINKED ISSUE #2004: [Clarification] Semantics & handling of duplicate query options values
Let's take the following examples:

1. /People?$select=Name,Name
2. /People?$orderby=Age asc,Age desc
3. /People?$expand=Relatives($filter=Married eq true),Relatives($filter=Married eq false)

Grammatically, the requests above are correct (per the [ABNF.](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/abnf/odata-abnf-construction-rules.txt) ).

Semantically however, they are confusing and/or full of contradictions. 
Is there any guidance about how OData services should handle duplicate expandItem/orderbyItem/etc. ?

The current example TripIn service does not error out, but only considers the last item. Not sure if that's intended or an oversight.
Ex: https://services.odata.org/V4/TripPinService/People?$select=UserName&$expand=Trips($select=TripId,Name),Trips($filter=Name eq 'Gradutaion trip'),Trips($filter=Name eq 'Trip in US')

ralfhandl:
Fixes #2004


odata-url-conventions/5 Query Options.md
```diff
@@ -2061,6 +2061,8 @@ Allowed system query options are
 [`$count`](#SystemQueryOptioncount), and
 [`$search`](#SystemQueryOptionsearch)
 for collection-valued navigation properties.
+The same system query option, irrespective of casing or whether or not it is prefixed with a `$`,
+MUST NOT be specified more than once for a collection-valued navigation property.
```
HeikoTheissen, line 2065: Why only collection-valued? Are duplicate `$compute` and `$expand` allowed after single-valued navigation properties?

HeikoTheissen (approved):

mikepizzo (approved):

PR #2011, opened 2024-10-10
Resource path in request body

HeikoTheissen:
Needs https://github.com/oasis-tcs/odata-abnf/pull/119


odata-url-conventions/4 Resource Path.md
```diff
@@ -1030,16 +1031,20 @@ $filter=LastName%20eq%20'P%26G'&$select=FirstName,LastName
 ```
 :::
 
+If the resource path is passed in the request body, it MUST be percent-encoded,
+come before any query options and be separated from them by a `?`.
```
ralfhandl, line 1035:
```suggestion
come before any query options, and be separated from them by a `?`.
```

ralfhandl: In other words: standard representation of a (relative) URL.

Can the resource path be a relative URL, and is it then resolved relative to the request URL?

Is only `<service-root>/$query` allowed, or can I use any URL suffixed with `/$query`?

ralfhandl (commented):
See comments

ralfhandl (dismissed):

HeikoTheissen:
As agreed in TC 2024-10-16.

PR #2013, opened 2024-10-16
Clarify ordering constraints of deleted entity in JSON payload

### LINKED ISSUE #1985: Clarify ordering of control information for deleted entity
According to [payload ordering constraints](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#PayloadOrderingConstraints), `@type` immediately follows `@context`, and other control information like `@id` follows `@type`:

> - If present, the context control information MUST be the first property in the JSON object.
> - The type control information, if present, MUST appear next in the JSON object.
> - The id and etag control information MUST appear before any property, property annotation, or property control information.

However, the section on [deleted entity](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#DeletedEntity) states that `@removed` should come immediately after `@context`:

> For ordered payloads, the control information `removed` MUST immediately follow the `context` control information, if present, otherwise it MUST be the first property in the deleted entity.

and later suggests that `@id` comes immediately after `@removed`.

> For ordered payloads, the control information `id`, if present, MUST immediately follow the control information `removed`.
 
I believe that we failed to update the section on ordered payloads for `@removed`, and the text for deleted entity is just missing `@type` all together, because originally there was no reason to specify the type of a deleted entity. Later we introduced the ability to include other properties on a deleted entity, which added the requirement to include `@type` for types derived from the expected type.
 
So, if we simply neglected to include `@removed` from ordering and `@type` for deleted entity, then you could argue the expected ordering is `@context`, `@removed`, `@type`, `@id`, or perhaps `@context`, `@type`, `@removed`, `@id`.

Whether the entity is deleted is arguably more important than its type, and is more similar to OData 4.0, in which the context conveyed whether or not it was deleted, so I would argue the first makes the most sense.

Proposal:
Clarify, in payload ordering constraints, that `@removed`, if present, should come after `@context` and before `@type`, if present.
Clarify, in deleted entity, that `@type`, if present, should come after `@removed` and before `@id`.

mikepizzo:
Fixes #1985


odata-json-format/4 Common Characteristics.md
```diff
@@ -164,6 +164,8 @@ constraints have to be met:
 
 - If present, the `context` control information MUST be the first
   property in the JSON object.
+- For 4.01 deleted entities, the `@removed` control information MUST appear
```
HeikoTheissen, line 167:
```suggestion
- For 4.01 deleted entities, the `removed` control information MUST appear 
```
Elsewhere you omit the leading `@`.

ralfhandl (dismissed):

HeikoTheissen (commented):
Section 4.5 demands the order: context, removed, type, [etag, ] id
Section 15.3. demands the order: [context, ] removed, [type, ] id
Can section 15.3 instead refer to section 4.5?

HeikoTheissen (dismissed):

ralfhandl (approved):

PR #2016, opened 2024-10-24
Use really simple identifiers

### LINKED ISSUE #375: Clarify support for supplementary plane characters in OData Identifiers
in [OData Common Schema Definition Language (CSDL) XML Representation Version 4.01 (oasis-open.org)](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_SimpleIdentifier) we define the rules for a legal identifier (for instance, a property name) in OData as follows:
> A simple identifier is a Unicode character sequence with the following restrictions:
> - It consists of at least one and at most 128 Unicode characters (code points).
> - The first character MUST be the underscore character (U+005F) or any character in the Unicode category “Letter (L)” or “Letter number (Nl)”.
> - The remaining characters MUST be the underscore character (U+005F) or any character in the Unicode category “Letter (L)”, “Letter number (Nl)”, “Decimal number (Nd)”, “Non-spacing mark (Mn)”, “Combining spacing mark (Mc)”, “Connector punctuation (Pc)”, and “Other, format (Cf)”.
> 
> Non-normatively speaking it starts with a letter or underscore, followed by at most 127 letters, underscores or digits.
> 

The Unicode Category L is defined as Lu | Ll | Lt | Lm | Lo.

Where Lo is "Other Letter".

Question: did we intent to include Surrogates (see [The Unicode Standard, Version 15.0)](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#G2630) (or, more properly, [supplementary plane characters](https://en.wikipedia.org/wiki/Plane_(Unicode))) as valid characters in an OData identifier?

The rules for valid property names were largely taken from programming languages in order to facilitate mapping between OData types and language objects. C#, for example, does not allow surrogates in property names.

Note that [ODATA-1348](https://issues.oasis-open.org/browse/ODATA-1348) tries to distinguish between characters and code points with regard to things like max string length, but this doesn't address character validity within an identifier.

### Proposal

We believe, according to the current rules, that characters from the supplementary plane would be included in Lo, so are valid in identifiers.

However, rather than call out characters that we don't recommend, we will leave that as an exercise to the reader and instead add a comment that, in order to support maximum interoperability, services SHOULD only use A-Z, a-z, 0-9, and underscore in identifiers.

Imported from [ODATA-1599](https://issues.oasis-open.org/browse/ODATA-1599)

ralfhandl:
Fixes #375


odata-csdl/15 Identifier and Path Values.md
```diff
@@ -28,6 +28,10 @@ restrictions:
 Non-normatively speaking it starts with a letter or underscore, followed
 by at most 127 letters, underscores or digits.
 
+For maximum interoperability services SHOULD use simple identifiers
+consisting only of latin letters, the underscore, and decimal numbers,
+that is match the pattern `^[_A-Za-z][_A-Za-z0-9]*$`.
```
HeikoTheissen, line 33:
```suggestion
that is, match the pattern `^[_A-Za-z][_A-Za-z0-9]*$`.
```

odata-csdl/15 Identifier and Path Values.md
```diff
@@ -28,6 +28,10 @@ restrictions:
 Non-normatively speaking it starts with a letter or underscore, followed
 by at most 127 letters, underscores or digits.
 
+For maximum interoperability services SHOULD use simple identifiers
+consisting only of latin letters, the underscore, and decimal numbers,
```
HeikoTheissen, line 32:
```suggestion
consisting only of letters from the ASCII charset, the underscore, and decimal digits,
```
The Latin alphabet did not have the letters J, U, W.

HeikoTheissen (changes requested):

odata-csdl/15 Identifier and Path Values.md
ralfhandl: We use Unicode terminology here, see https://www.compart.com/en/unicode/U+004A, https://www.compart.com/en/unicode/U+0055, https://www.compart.com/en/unicode/U+0057

HeikoTheissen (approved):

mikepizzo (approved):

mikepizzo (approved):

PR #2017, opened 2024-10-24
325 byte order mark in text/plain

### LINKED ISSUE #325: Clarify use of byte-order mark (BOM) in text/plain responses
The specification is currently unclear on the use of byte-order marks in "plain text" responses to e.g. `/$count` segments or `/$value` requests to numeric properties.

### Proposal

OData services [REALLY SHOULD NOT](https://tools.ietf.org/html/rfc6919#section-3) send a byte-order mark in responses with `Content-Type: text/plain`.

Clients SHOULD be aware that it is returned by some services.

Imported from [ODATA-1426](https://issues.oasis-open.org/browse/ODATA-1426)

ralfhandl:
Fixes #325


odata-protocol/11 Data Service Requests.md
```diff
@@ -255,22 +255,24 @@ annotation (see [OData-VocCore](#ODataVocCore)) if this
...
+
+Responses for type `Edm.String` MAY use the `charset` format parameter to specify the
+character set used for representing the string value; omission of the `charset` parameter implies the default character set `US-ASCII`.
+
+Responses for the other primitive types follow the rules `booleanValue`, `byteValue`,
 `dateValue`, `dateTimeOffsetValue`, `decimalValue`, `doubleValue`,
 `durationValue`, `enumValue`, `guidValue`, `int16Value`, `int32Value`,
 `int64Value`, `sbyteValue`, `singleValue`, and `timeOfDayValue` in
-[OData-ABNF](#ODataABNF).
+[OData-ABNF](#ODataABNF); they MUST NOT use the `charset` parameter which implies the default character set `US-ASCII`.
```
HeikoTheissen, line 275:
```suggestion
[OData-ABNF](#ODataABNF); they MUST NOT use the `charset` parameter since they use the default character set `US-ASCII`.
```
Sounded as if "the `charset` parameter implies the default character set `US-ASCII`".

odata-protocol/11 Data Service Requests.md
```diff
@@ -1145,8 +1147,10 @@ On success, the response body MUST contain the exact count of items
 matching the request after applying any
 [`$filter`](#SystemQueryOptionfilter) or
 [`$search`](#SystemQueryOptionsearch) system query options, formatted as
-a simple primitive integer value with media type `text/plain`. Clients
-SHOULD NOT combine the system query options
+a simple primitive integer value with media type `text/plain` without `charset` parameter,
```
HeikoTheissen, line 1150: I may be paranoid, but it is specified that the integer must be given in base-10 notation? (Javascript interprets `010` as octal notation for the number eight.)

HeikoTheissen (changes requested):

odata-protocol/11 Data Service Requests.md
ralfhandl: https://www.rfc-editor.org/rfc/rfc2046.html#section-4.1.2 states
> The default character set, which must be assumed in the absence of a charset parameter, is US-ASCII.

ralfhandl: Maybe I should add a reference.

odata-protocol/11 Data Service Requests.md
```diff
@@ -255,22 +255,25 @@ annotation (see [OData-VocCore](#ODataVocCore)) if this
...
+
+Responses for type `Edm.String` MAY use the `charset` format parameter to specify the
+character set used for representing the string value; omission of the `charset` parameter implies the default character set `US-ASCII`.
+
+Responses for the other primitive types follow the rules `booleanValue`, `byteValue`,
 `dateValue`, `dateTimeOffsetValue`, `decimalValue`, `doubleValue`,
 `durationValue`, `enumValue`, `guidValue`, `int16Value`, `int32Value`,
 `int64Value`, `sbyteValue`, `singleValue`, and `timeOfDayValue` in
-[OData-ABNF](#ODataABNF).
+[OData-ABNF](#ODataABNF); they MUST NOT use the `charset` parameter which implies the default character set `US-ASCII`,
```
HeikoTheissen, line 275:
```suggestion
[OData-ABNF](#ODataABNF); they MUST NOT use the `charset` parameter, this implies the default character set `US-ASCII`,
```

odata-protocol/11 Data Service Requests.md
```diff
@@ -1149,7 +1149,7 @@ matching the request after applying any
 [`$filter`](#SystemQueryOptionfilter) or
 [`$search`](#SystemQueryOptionsearch) system query options, formatted as
 a simple primitive integer value with media type `text/plain` without `charset` parameter,
-and the response body consisting only of octets `0x30` to `0x39`.
+and the response body following rule `int64value` without the sign, see [OData-ABNF](#ODataABNF).
```
HeikoTheissen, line 1152:
```suggestion
and the response body following rule `int64Value` without the sign, see [OData-ABNF](#ODataABNF).
```

HeikoTheissen (changes requested):

odata-protocol/11 Data Service Requests.md
ralfhandl: We could do that. On the other hand the [control information count (odata.count)](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#ControlInformationcountodatacount) is an `Edm.Int64` value, and I assumed the same to be true for the `/$count` value.

odata-protocol/11 Data Service Requests.md
```diff
@@ -1145,8 +1148,10 @@ On success, the response body MUST contain the exact count of items
 matching the request after applying any
 [`$filter`](#SystemQueryOptionfilter) or
 [`$search`](#SystemQueryOptionsearch) system query options, formatted as
-a simple primitive integer value with media type `text/plain`. Clients
-SHOULD NOT combine the system query options
+a simple primitive integer value with media type `text/plain` without `charset` parameter,
+and the response body following rule `1*DIGIT`.
```
HeikoTheissen, line 1152:
```suggestion
and the response body following rule `1*DIGIT`, see [OData-ABNF](#ODataABNF).
```
To make clear what `DIGIT` means.

HeikoTheissen (changes requested):

HeikoTheissen (approved):

Hubert-Heijkers (approved):
These changes are fine but, unfortunately, only apply to text/plain (which in my mind never were an issue as that's US-ASCII anyway, to being explicit is a good clarification).

Hubert-Heijkers:
I realize the title of this issue being resolved says "use of byte-order mark (BOM) in **text/plain** responses" but, as I've expressed originally in a comment on the Jira issue, my interest in this topic of BOMs was more related to responses that actually use a Unicode charset, specifically UTF-8, and as such I'd hoped that we could also make explicit that, based on the remarks wrt the Byte-Order Mark in [RFC 3629](https://tools.ietf.org/html/rfc3629#section-6) that OData services MUST NOT (SHOULD NOT if that would break existing rules) send a byte-order mark in UTF-8 responses (read: irrespective of content-type, with UTF-8 charset implied or explicitly specified using charset parameter).

Should I create a separate issue or can we entertain fixing that as part of this issue as well?

PR #2019, opened 2024-10-25
`null` result with instance annotations

### LINKED ISSUE #2018: Instance annotations and control information for a `null` result
According to [OData-JSON, section 11](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#IndividualPropertyorOperationResponse), an individual property or operation response with value `null` must be returned as 204 No Content.

In [version 4.0](https://docs.oasis-open.org/odata/odata-json-format/v4.0/csprd01/odata-json-format-v4.0-csprd01.html#_Toc355172944) it was still returned as 200 OK with
```json
{"value": null}
```

I would like to re-allow the former format (as an alternative to 204 No Content), because it is the only way to add instance annotations or control information to the `null` result.

Use cases:
- Instance annotation containing messages that occurred during the operation (and which may explain why the result is `null`)
- [`@deltaLink`](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#ControlInformationdeltaLinkodatadeltaLink) not only for "sets of results", but also for single-valued results. Could be used to obtain changes to individual properties or related entities, and must then also be issued for a `null` result.

HeikoTheissen:
Fixes #2018

### DIFF
--- a/odata-json-format/10 Media Entity.md
+++ b/odata-json-format/10 Media Entity.md
@@ -33,19 +33,24 @@ An individual property or operation response is represented as a JSON
 object.
 
 A single-valued property or operation response that has the
-`null` value does not have a representation; see
+`null` value and carries no [control information](#ControlInformation)
+or [instance annotations](#InstanceAnnotations)
+does not have a representation; see
 [OData-Protocol](#ODataProtocol).
 
 A property or operation response that is of a primitive type is
-represented as an object with a single name/value pair, whose name is
+represented as an object with a name/value pair whose name is
 `value` and whose value is a [primitive
-value](#PrimitiveValue).
+value](#PrimitiveValue) or `null`.
 
 A property or operation response that is of complex type is represented
-as a [complex value](#ComplexValue).
+as a [complex value](#ComplexValue). If the value is `null`, the JSON object
+consists of name/value pairs for the `@context` and the
+[control information](#ControlInformation)
+and [instance annotations](#InstanceAnnotations) only.
 
 A property or operation response that is of a collection type is
-represented as an object with a single name/value pair whose name is
+represented as an object with a name/value pair whose name is
 `value`. Its value is the JSON representation of a
 [collection of complex type values](#CollectionofComplexValues) or
 [collection of primitive values](#CollectionofPrimitiveValues).
@@ -100,6 +105,20 @@ Example ##ex: empty collection of complex values
 {
   "@context": "http://host/service/$metadata#Collection(Model.Address)",
   "value": []
+}
+```
+:::
+
+::: example
+Example ##ex: `null` value accompanied by an instance annotation
+```json
+{
+  "@context": "http://host/service/$metadata#Model.Address",
+  "@Core.Messages": [{
+    "code": "EPOSTCODE",
+    "message": "Could not resolve postcode XYZ",
+    "severity": "error"
+  }]
 }
 ```
 :::
--- a/docs/odata-json-format/odata-json-format.html
+++ b/docs/odata-json-format/odata-json-format.html
@@ -1269,10 +1269,10 @@ http://host/service/Customers(&#39;ALFKI&#39;)/Orders</code></pre>
 <h1 id="11-individual-property-or-operation-response"><a id="IndividualPropertyorOperationResponse" href="#IndividualPropertyorOperationResponse">11 Individual Property or Operation Response</a></h1>
 </summary>
 <p>An individual property or operation response is represented as a JSON object.</p>
-<p>A single-valued property or operation response that has the <code>null</code> value does not have a representation; see <a href="#ODataProtocol">OData-Protocol</a>.</p>
-<p>A property or operation response that is of a primitive type is represented as an object with a single name/value pair, whose name is <code>value</code> and whose value is a <a href="#PrimitiveValue">primitive value</a>.</p>
-<p>A property or operation response that is of complex type is represented as a <a href="#ComplexValue">complex value</a>.</p>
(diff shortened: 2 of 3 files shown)


odata-json-format/10 Media Entity.md
```diff
@@ -33,19 +33,24 @@ An individual property or operation response is represented as a JSON
...
 A property or operation response that is of a primitive type is
-represented as an object with a single name/value pair, whose name is
+represented as an object with a name/value pair whose name is
 `value` and whose value is a [primitive
-value](#PrimitiveValue).
+value](#PrimitiveValue) or `null`.
 
 A property or operation response that is of complex type is represented
-as a [complex value](#ComplexValue).
+as a [complex value](#ComplexValue). If the value is `null`, the JSON object
```
HeikoTheissen, line 47: Are entity-valued results covered by this paragraph on complex values as well? They are not covered elsewhere.

HeikoTheissen, line 47: Are entity-valued results covered by this paragraph on complex values as well? They are not covered elsewhere.

ralfhandl: How do I distinguish `null` from a complex instance without properties?

HeikoTheissen: > Entity-valued responses are covered in [6 Entity](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#Entity)

This does not mention entities whose value is `null` (a nullable single-valued navigation property, for example).

odata-json-format/10 Media Entity.md
```diff
@@ -33,19 +33,24 @@ An individual property or operation response is represented as a JSON
 object.
 
 A single-valued property or operation response that has the
-`null` value does not have a representation; see
-[OData-Protocol](#ODataProtocol).
+`null` value does not have a representation provided that it carries no
+[control information](#ControlInformation) other than
+[`context`](#ControlInformationcontextodatacontext)
+and no [instance annotations](#InstanceAnnotations); see [OData-Protocol](#ODataProtocol).
```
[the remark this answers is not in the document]
HeikoTheissen: To be done later, once the changes to [OData-JSON] are agreed upon.

odata-json-format/10 Media Entity.md
ralfhandl: > This does not mention entities whose value is `null` (a nullable single-valued navigation property, for example).

That is in [8.3 Expanded Navigation Property](https://docs.oasis-open.org/odata/odata-json-format/v4.02/csd01/odata-json-format-v4.02-csd01.html#ExpandedNavigationProperty):

> If at most one entity can be related, the value is the representation of the related entity, or `null` if no entity is currently related.

HeikoTheissen: I wasn't talking about _expansion_ but about requesting a _related_ entity, which is absent, like `GET Employees('CEO')/Manager`. I would expect 204 No Content here (unless there is control information or instance annotations).

Or about invoking an operation with entity-valued return type, if that returns null.

ralfhandl: > requesting a _related_ entity, which is absent, like `GET Employees('CEO')/Manager`

That is covered in [11.2.7 Requesting Related Entities](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#RequestingRelatedEntities):

> If the relationship terminates on a single entity, the response MUST be the format-specific representation of the related single entity. If no entity is related, the service returns [204 No Content](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#ResponseCode204NoContent).

"No Content" means no representation, so this is not mentioned in the JSON Format.

odata-json-format/10 Media Entity.md
```diff
@@ -33,19 +33,24 @@ An individual property or operation response is represented as a JSON
 object.
 
 A single-valued property or operation response that has the
-`null` value does not have a representation; see
-[OData-Protocol](#ODataProtocol).
+`null` value does not have a representation provided that it carries no
```
HeikoTheissen, line 36: What does this mean when this section is applied recursively, while representing the properties of an entity?

Section 6 says "each structural property is represented as a name/value pair", and I understand that the value is the representation of the property. But now this section here says that the property "has no representation".

ralfhandl: This section describes the full response body of an individual property or operation response. It is not intended to be applied recursively.

ralfhandl: Similar to collection responses: they are also not intended for recursive application, otherwise instead of `"coll":[...]` we would see `"coll":{"value":[...]}`

odata-json-format/10 Media Entity.md
```diff
@@ -37,15 +37,15 @@ A single-valued property or operation response that has the
 [OData-Protocol](#ODataProtocol).
 
 A property or operation response that is of a primitive type is
-represented as an object with a single name/value pair, whose name is
+represented as an object with a name/value pair whose name is
```
HeikoTheissen, line 40: Deleted "single", because the JSON object can also contain control information and instance annotations.

odata-json-format/10 Media Entity.md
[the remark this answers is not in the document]
HeikoTheissen: I have now moved almost all changes to [OData-Protocol]. And that's where it belongs, I now see.

ralfhandl (approved):

PR #2020, opened 2024-10-29
Add Node 22

ralfhandl:
Node 22 is now current, Node 20 is active LTS, see https://github.com/nodejs/release?tab=readme-ov-file#release-schedule

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20.x, 22.x]
+        node-version: [22.x]
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


HeikoTheissen (changes requested):
No need to run the build twice.

ralfhandl:
Removing `20.x` before having the branch protection rule changed makes this PR unmergeable:

image

Not sure whether this will repair itself once the branch protection rule is changed. Let's wait and see.

HeikoTheissen (dismissed):

.github/workflows/nodejs.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [20.x]
+        node-version: [22.x]
```
ralfhandl, line 11:
```suggestion
        node-version: [20.x, 22.x]
```

HeikoTheissen (approved):

PR #2048, opened 2025-02-11
Prepare OData-Data-Agg for OASIS standard

HeikoTheissen:


odata-data-aggregation-ext/3.2 Basic Aggregation.md
```diff
@@ -370,6 +370,8 @@ The output set of the transformation $\Pi_G(s)$ is in one-to-one correspondence
    - Otherwise, the behavior is undefined. (Such cases never occur when $\Pi_G(s)$ is used in this document.)
 3. Return $u$.
 
+Instances in the output set of $\Pi_G(s)$ that contain the key of an entity MAY retain that entity's applicable control information [#OData-JSON#ControlInformation].
```
mikepizzo, line 373: something like (feel free to edit as appropriate)
```suggestion
Instances in the output set of $\Pi_G(s)$ that contain the key of an entity retain the metadata associated with the entity instance, such as id and etag.
```
and *possibly* state that this information *may* be returned in the final response according to the format-specific conventions.

PR #2049, opened 2025-02-13
Omission of collection-valued parameters/properties

### LINKED ISSUE #2045: Omission of collection-valued action parameters
[OData-JSON, section 18](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#ActionInvocation) states
> Non-binding parameters that are nullable ... MAY be omitted from the request body. ... it MUST be interpreted as having the null value.

This statement cannot apply to collection-valued parameters, because they can at most be interpreted as having an empty collection as value. Is this meant here?
- If yes, can an empty collection be omitted even if non-nullable?
- If not, a collection-valued parameter must always be included even if empty.

Conversely, can an empty array be omitted in a response, similar to `omit-values=nulls`?

HeikoTheissen:
Fixes #2045


odata-json-format/16 Bound Function.md
```diff
@@ -337,6 +337,11 @@ how to interpret the omitted parameter. Note: a nullable non-binding
 parameter is equivalent to being annotated as optional with a default
 value of `null`.
 
+Non-binding collection-valued parameters MAY be omitted from the request body regardless
+of nullability or optionality. Unless a `Core.OptionalValue` with `DefaultValue` is annotated,
+the interpretation of such an omitted parameter is up to the service,
+this includes assuming an empty collection or reporting an error.
```
HeikoTheissen, line 343: Note: This allows a service to treat an absent mandatory parameter of type `Collection(Edm.Int16)` like `[1,2]`.

ralfhandl (approved):

odata-json-format/16 Bound Function.md
ralfhandl, line 343: Mention this in the "Changes" section

odata-protocol/8 Header Fields.md
```diff
@@ -599,17 +599,19 @@ SHOULD be used.
...
 response payload. Valid values are `nulls` or `defaults`.
 
-If `nulls` is specified, then the service MAY omit properties containing
+If `nulls` is specified, then the service MAY omit single-valued properties having
 null values from the response, in which case it MUST specify the
 `Preference-Applied` response header with `omit-values=nulls`.
 
-If `defaults` is specified, then the service MAY omit properties
+If `defaults` is specified, then the service MAY omit single-valued properties
 containing default values from the response, including nulls for
```
ralfhandl, line 607: "containing" -> "having"

odata-json-format/16 Bound Function.md
```diff
@@ -337,6 +337,11 @@ how to interpret the omitted parameter. Note: a nullable non-binding
 parameter is equivalent to being annotated as optional with a default
 value of `null`.
 
+Non-binding collection-valued parameters MAY be omitted from the request body regardless
+of nullability or optionality. Unless a `Core.OptionalValue` with `DefaultValue` is annotated,
```
ralfhandl, line 341: DefaultValue has type String, so it only works for primitive single-valued :-(

ralfhandl: We could extend the term `Core.OptionalParameter` with `AnyDefaultValue` or `NonPrimitiveDefaultValue` or similar of type `Edm.Untyped`.

ralfhandl (approved):

ralfhandl (dismissed):

ralfhandl (approved):

odata-json-format/16 Bound Function.md
```diff
@@ -337,6 +337,11 @@ how to interpret the omitted parameter. Note: a nullable non-binding
 parameter is equivalent to being annotated as optional with a default
 value of `null`.
 
+Non-binding collection-valued parameters MAY be omitted from the request body regardless
+of nullability or optionality. The interpretation of such an omitted parameter is up to the service,
```
mikepizzo, line 341: "May be omitted" and "reporting an error" seem in conflict.  Can we remove the first part, and just say something like "If collection-valued parameters are omitted from the request, the interpretation is up to the service,..."

mikepizzo (approved):

PR #2058, opened 2025-03-01
Effective absence of dynamic expressions

### LINKED ISSUE #2057: Annotation default values if dynamic expression is null
Consider the following annotations:
```json
"@Capabilities.ComputeSupported": {"$Path": "NavigationProperty/BooleanProperty"},
"@Capabilities.SelectSupport": {"ComputeSupported": {"$Path": "NavigationProperty/BooleanProperty"}}
```

What are the annotation values if the `NavigationProperty` is null?

`SelectSupport/ComputeSupported` is effectively not given, because null is not an allowed value. Hence its default value `false` applies.

`ComputeSupported` is effectively not given, because null is not an allowed value. But does that mean that the default value `true` applies? In other words and in XML parlance, is this equivalent to
- `<Annotation Term="Capabilities.ComputedSupported" />` (in which case the default value `true` would apply) or to
- no `<Annotation>` at all, in which no default value would apply?

HeikoTheissen:
Fixes #2057


ralfhandl (dismissed):

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -3161,6 +3209,15 @@ the base term or its base term etc. need not be specified again.
 For collection-valued properties the absence of a property value
 expression is equivalent to specifying an empty collection as its value.
 
+If a dynamically provided property value is not acceptable for the type of the property or
+its [facets](#TypeFacets), does not meet the constraints imposed by
+its [nullability](#Nullable) or by
+its annotations from the Validation vocabulary [OData-VocValidation](#ODataVocValidation),
+no value is effectively provided for the property.
+In this case the client MAY apply the rules for absent properties:
```
HeikoTheissen, line 3217: By using MAY, we do not force clients to treat such "errors" in the same way as actually omitted properties.

ralfhandl (approved):

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -1309,6 +1309,18 @@ Example ##ex:
...
+its [facets](#TypeFacets), does not meet the constraints imposed by its
+: varjson
+`$Nullable` property
+:
+: varxml
+`Nullable` attribute
+:
+or by its
+annotations from the Validation vocabulary [OData-VocValidation](#ODataVocValidation),
+then the term is effectively not applied (see [example ##termdefault]).
```
mikepizzo, line 1322: I think we want to say the client can make no assumptions, because the server tried to say something about the term but it was not valid.

odata-csdl/14 Vocabulary and Annotation.md
```diff
@@ -1309,6 +1309,19 @@ Example ##ex:
 Dynamic expressions allow assigning a calculated value to an applied
 term.
 
+If a calculated value is not acceptable for the type of the term or
+its [facets](#TypeFacets), does not meet the constraints imposed by its
+: varjson
+`$Nullable` property
```
ralfhandl, line 1315: We typically use "member" or "name/value pair" when speaking of JSON objects, following https://datatracker.ietf.org/doc/html/rfc8259#section-4.

```suggestion
`$Nullable` member
```

ralfhandl (commented):
+1, just avoid "property" when talking about JSON object members

ralfhandl (approved):

PR #2059, opened 2025-03-03
Specify how entities returned by $compute are explicitly included in the response

### LINKED ISSUE #2047: Specify how entities returned by $compute are explicitly included in the response
$compute takes a common expression and assigns it an alias that is, by default, included in the response.

We state the following:
"Computed properties SHOULD be included as dynamic properties in the result and MUST be included if $select is specified with the computed property name, or star (*)."

However, since a common expression could return an entity type, should $expand be used to explicitly include in a result?

### LINKED ISSUE #2063: Add more information on dynamic properties
Provide more explanation on dynamic properties, for example
- need not exist on all instances
- may have different type on different instances, even alternate between stream, entity, structural, collection, ...
- can be specified in both `$select` and `$expand` and are interpreted per instance according to their type
  - entity (collection) in `$select`: include control information
  - stream in `$select` include control information
  - entity (collection) or stream in `$expand`: include data
  - structural in `$expand`: ignore
- ...

A possible location for this is Protocol, 3 Data Model, where dynamic properties are mentioned first.

Link to this information from section on `$compute`.

ralfhandl:
Fixes #2047
Fixes #2063


odata-protocol/11 Data Service Requests.md
```diff
@@ -555,9 +555,9 @@ properties that can be used in a [`$select`](#SystemQueryOptionselect)
 or within a [`$filter`](#SystemQueryOptionfilter) or
 [`$orderby`](#SystemQueryOptionorderby) expression.
 
-Computed properties SHOULD be included as dynamic properties in the
-result and MUST be included if `$select` is specified with the computed
-property name, or star (`*`).
+Computed properties SHOULD be included as dynamic properties in the result.
+Computed properties whose value is a stream, an entity, or a collection of entities MUST be included if `$expand` is specified with the computed property name or star (`*`).
```
HeikoTheissen, line 559: `$expand=*` does not expand stream properties, they must be explicitly named.

HeikoTheissen (changes requested):

odata-protocol/11 Data Service Requests.md
mikepizzo: Discussion 2025-3-12: Expanding streams was added after expand=*. The server is always allowed to expand more than requested, but requiring expand=* to include streams is arguably not backward compatible.

Most compatible solution would be to continue the current definition of expand=* not including streams; streams must be explicitly enumerated.

odata-protocol/11 Data Service Requests.md
```diff
@@ -555,9 +555,10 @@ properties that can be used in a [`$select`](#SystemQueryOptionselect)
 or within a [`$filter`](#SystemQueryOptionfilter) or
 [`$orderby`](#SystemQueryOptionorderby) expression.
 
-Computed properties SHOULD be included as dynamic properties in the
-result and MUST be included if `$select` is specified with the computed
-property name, or star (`*`).
+Computed properties SHOULD be included as dynamic properties in the result.
+Computed properties whose value is a stream, an entity, or a collection of entities MUST be included if `$expand` is specified with the computed property name.
+Computed properties whose value is an entity or a collection of entities MUST be included if `$expand` is specified with star (`*`).
+Computed properties with other values MUST be included if `$select` is specified with the computed property name or star (`*`).
```
mikepizzo, line 561:
```suggestion
Computed properties with types other than stream, entity, or collection of entity MUST be included if `$select` is specified with the computed property name or star (`*`).
```

HeikoTheissen (approved):

odata-protocol/11 Data Service Requests.md
```diff
@@ -551,13 +551,17 @@ GET http://host/service/Employees?$expand=Model.Manager/DirectReports($levels=4)
...
-property name, or star (`*`).
+properties that can be used in [`$expand`](#SystemQueryOptionexpand),
+[`$select`](#SystemQueryOptionselect),
+[`$filter`](#SystemQueryOptionfilter), or
+[`$orderby`](#SystemQueryOptionorderby).
+They are treated according to their instance-specific type, with the addition that they are silently ignored in `$expand` if the instance-specific value is neither a stream, an entity, or a collection of entities.
+But if the service can infer the that a computed property's type makes it unsuitable for `$expand` independently of the underlying data, it MUST reject the request
+The inferral can be based on, for example, the
+declared type of a property or the type of a literal value that occurs in the
+expression in the compute instruction.
```
ralfhandl, line 562:
```suggestion
compute instruction.
```

HeikoTheissen (dismissed):

odata-protocol/11 Data Service Requests.md
```diff
@@ -551,13 +551,17 @@ GET http://host/service/Employees?$expand=Model.Manager/DirectReports($levels=4)
...
+[`$select`](#SystemQueryOptionselect),
+[`$filter`](#SystemQueryOptionfilter), or
+[`$orderby`](#SystemQueryOptionorderby).
+They are treated according to their instance-specific type, with the addition that they are silently ignored in `$expand` if the instance-specific value is neither a stream, an entity, or a collection of entities.
+If the service can infer that a computed property's type makes it unsuitable for `$expand` independently of the underlying data, it MUST reject the request.
+The inferral can be based on, for example, the
+declared type of a property or the type of a literal value that occurs in the
+compute instruction.
+
+Computed properties SHOULD be included as dynamic properties in the result without being explicitly mentioned in `$expand`or `$select`, or implied by star (`*`).
```
mikepizzo, line 564:
```suggestion
The `$compute` system query option allows clients to define computed properties that can be used in [`$expand`](#SystemQueryOptionexpand),
[`$select`](#SystemQueryOptionselect),
[`$filter`](#SystemQueryOptionfilter), or
[`$orderby`](#SystemQueryOptionorderby). Computed properties are expanded or selected according to their instance-specific type. A computed property is ignored by `$expand` if its instance-specific value is neither a stream, an entity, nor a collection of entities, and the service cannot determine this based on the compute expression.

Computed properties SHOULD be included as dynamic properties in the result without being explicitly mentioned in `$expand` or `$select`, or implied by star (`*`).
```

mikepizzo (approved):

PR #2062, opened 2025-03-06
Insert a singleton

### LINKED ISSUE #2060: Nullable single-valued navigation property or singleton: update or upsert?
[OData-Protocol, section 11.4.4](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#UpsertanEntity) allows upserts for nullable
- single-valued navigation properties under certain conditions
- singletons generally.

How does an upsert differ from an update in these cases? The thinking seems to be "a change from null to non-null is treated like an insert". I suggest to make this explicit:

§1. The introductory sentence says
> An upsert occurs when the client sends an [update request](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#UpdateanEntity) to a valid URL that identifies a single entity that does not yet exist. In this case the service MUST handle the request as a [create entity request](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#CreateanEntity) or fail the request altogether.

Translated for our case, "does not yet exist" means "is null", and (contrary to [section 11.4.2](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#CreateanEntity)) the "create entity request" does not
> create an entity in a collection.

§2. The last two paragraphs explain how the `If-(None-)Match` header can be used to distinguish between insert and update. Translated for our case "insert" means "change from null to non-null" (or even to null?), and "update" means "change from non-null to any value".

§3. Clarify the meaning of `Upsertable` and `Updatable` in the capabilities annotation:
https://github.com/oasis-tcs/odata-vocabularies/blob/229c2cd0aca61fcfb9ca66f437ad144ad227a766/vocabularies/Org.OData.Capabilities.V1.xml#L712-L717

- What does the combination of `Upsertable = false` and `Updatable = true` (the default values) mean in our single-valued case? Is a change from null to non-null then allowed?
- The combination of `Upsertable = true` and `Updatable = false` makes no sense for entities in a collection. But for our single-valued case this would translate into "PATCH request allowed only if current value is null". The client could enforce this with `If-None-Match: *`. Correct?

HeikoTheissen:
Fixes #2060


ralfhandl (dismissed):

odata-protocol/11.4 Data Modification.md
```diff
@@ -159,7 +159,10 @@ related entity, those related entities MUST be included either as
 references to existing entities or as content for new related entities.
 
 An entity may also be created as the result of an
```
HeikoTheissen, line 161: Does this cover the case of a nullable singleton? Singletons are not entities, are they?

odata-protocol/11.4 Data Modification.md
```diff
@@ -159,7 +159,10 @@ related entity, those related entities MUST be included either as
...
-[Upsert](#UpsertanEntity) operation.
+[Upsert](#UpsertanEntity) operation, which is then said to be "treated as an insert".
+This includes the case where a nullable singleton
+or single-valued navigation property is changed from null to any value
+(even the redundant change from null to null).
```
ralfhandl, line 165: Why mention the change from null to null? Does this statement imply that the service cannot simply do nothing when it receives this redundant "no-op" change?

odata-protocol/11.4 Data Modification.md
```diff
@@ -364,7 +367,11 @@ property `$value`.
 Updating a dependent property that is tied to a key property of the
 principal entity through a referential constraint updates the
 relationship to point to the entity with the specified key value. If
-there is no such entity, the update fails.
+the entity set for that entity cannot be determined or does not contain
+such an entity, the update fails. The entity set can be determined, among
+other possibilities, by containment navigation properties or navigation property bindings in the
+service metadata, or by `@context` or `@id` control information in the request
+payload.
```
ralfhandl, line 374:
```suggestion
other possibilities, by containment or navigation property binding,
or by control information in the request payload.
```
I'd prefer to keep specifics of the JSON Format out of the Protocol specification.

HeikoTheissen:
We agreed yesterday that if I assign an existing customer 12 to an existing order 23 that had no customer assigned before
```
PATCH Orders(23)/Customer

{"CustomerID": "12"}
```
that counts as an update, not an upsert. I want to make sure this is not treated as an insert (no customer is created), can I include a header `If-Match: *`?
> To ensure that an update request is not treated as an insert, the client MAY specify an [If-Match](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#HeaderIfMatch) header in the update request.

That would be counter-intuitive, because if `Orders(23)/Customer` is `null` before the update, it does _not_ match `*`.

odata-protocol/11.4 Data Modification.md
```diff
@@ -657,17 +660,20 @@ the request.
...
-Upserts to single-valued navigation properties are possible for
-
-- containment navigation properties,
-- non-containment navigation properties with a navigation property binding, or
-- payloads including a context URL specifying the entity set or
-  contained collection of entities in which the new entity is to be created.
+Upserts are possible if the canonical collection or canonical singleton
+(as defined in [section ##ContextURL]) can be determined for the newly-created entity.
+Among other possibilities, this is the case if
+1. the URL contains the name of an entity set or singleton
```
HeikoTheissen, line 672: More precision needed? "URL contains a segment consisting of the name of an entity set or singleton"

odata-protocol/11.4 Data Modification.md
```diff
@@ -680,23 +689,45 @@ with the specified media stream if it does not already exist, otherwise the
...
+  instead of newly creating the addressed entity because the [etag](#UseofETagsforAvoidingUpdateConflicts)
+  does not match
+- contains an [`If-None-Match: *`](#HeaderIfNoneMatch) header, the service MUST fail the request
+  if it would update an existing entity in the absence of the header.
+
+In other words, the `If-`(`None-`)`Match` header distinguishes between update and upsert
+in these cases.
+
+If the URL of an update request does not fall into case 1, 2, or 3 given above, it ends
+with a non-containment navigation property, optionally followed by a key predicate.
```
HeikoTheissen, line 714: In this case, the client may not care whether the operation is an update or upsert, and hence may not want to evaluate both `Updatable` and `Upsertable` in the capabilities annotations.

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,20 +672,22 @@ operations.
 On failure, the service MUST NOT apply any of the changes specified in
 the request.
 
-### ##subsubsec Upsert an Entity
+#### ##subsubsubsec Upsert an Entity
```
ralfhandl, line 675: Given that the Update section requires failing the request when the addressed entity does not exist, I find it confusing to make this now a subsection.

ralfhandl (changes requested):
I think we need to restructure these two sections, for example putting Upsert as the general contract first and making pure Update a special case that only triggers if the client sends `If-Match:*`.

odata-protocol/11.4 Data Modification.md
```diff
@@ -364,7 +379,10 @@ property `$value`.
 Updating a dependent property that is tied to a key property of the
 principal entity through a referential constraint updates the
 relationship to point to the entity with the specified key value. If
-there is no such entity, the update fails.
+the entity set for that entity cannot be determined or does not contain
```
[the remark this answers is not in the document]
HeikoTheissen: This is already excluded because singletons have no key.

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,20 +672,22 @@ operations.
...
 
-### ##subsubsec Upsert an Entity
+#### ##subsubsubsec Upsert an Entity
 
-An upsert occurs when the client sends an [update
-request](#UpdateanEntity) to a valid URL that identifies a single entity
-that does not yet exist. In this case the service MUST handle the
-request as a [create entity request](#CreateanEntity) or fail the
-request altogether.
+An upsert occurs when an [update request](#UpdateanEntity) newly creates the addressed entity.
```
ralfhandl, line 677: An upsert is a request that updates the target if it exists, and inserts it if it does not yet exist. This is how [HTTP defines PUT](https://www.rfc-editor.org/rfc/rfc9110.html#name-put), and similar to the [MODIFY statement in ABAP](https://help.sap.com/doc/abapdocu_751_index_htm/7.51/de-DE/abapmodify_dbtab.htm).

The client doesn't know or care whether the target already exists, and the server does the needful.

If the client cares, it adds `If-Match: *` or `If-None-Match: *` to the request.

"Upsert" is the *client intention*, and "update or insert" is the *server reaction*.

HeikoTheissen: Let's decide the terminology in the TC.
What|Heiko|Ralf
---|---|---
client intention|update|upsert
if entity exists|no upsert|update
if entity is created|upsert|insert

odata-protocol/11.4 Data Modification.md
HeikoTheissen, line 1: Wording: URL of a collection/entity vs. resource path of a collection/entity

odata-protocol/11.4 Data Modification.md
ralfhandl: https://en.wiktionary.org/wiki/upsert

https://wiki.postgresql.org/wiki/UPSERT

https://www.cockroachlabs.com/blog/sql-upsert/

odata-protocol/11.4 Data Modification.md
```diff
@@ -148,6 +148,21 @@ to that collection's URL. The `POST` body MUST contain a single valid
 representation of an entity of the declared target entity type,
 or one of its derived types.
 
+If the service is unable to determine the canonical collection (as defined in [section ##ContextURL])
+for the entity, it MUST fail the request.
+This determination is possible, for example, if the collection's URL
+- consists of an entity set
+- ends with a containment navigation property
+- ends with a non-containment navigation property with a navigation property binding
```
HeikoTheissen, line 156: Is the navigation property binding really sufficient? Given
```xml
<EntitySet Name="Orders">
  <NavigationPropertyBinding Path="DeepLink" Target="ns.container/Orders/Items/Subitems" />
</EntitySet>
```
how can a `POST Orders(23)/DeepLink` determine the `x` in
```
Location: http://server/Orders(23)/Items(x)/Subitems(y)
```
for the newly-created entity? (`y` is generated by the service.)

docs/odata-protocol/odata-protocol.md
```diff
@@ -4722,25 +4743,57 @@ with the specified media stream if it does not already exist, otherwise the
...
+necessarily equivalent to the upsert distinction (see also
+[example 85](#specialization)). If the request
+- contains an `If-Match: *` header, the service MUST fail the request
+  if the navigation property is null
+- contains an `If-Match` header with an ETag as value, the service fails the request
+  if the navigation property is null because the ETag does not match
+- contains an `If-None-Match: *` header, the service MUST fail the request
+  if the navigation property is not null.
+
+Note that if the service does not fail here, it may update the navigation property to point to
```
uhlmannm, line 4775: Which leaves the question whether to check [Updatable](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md#UpdateRestrictionsType) or Upsertable before sending such a request.

odata-protocol/11.4 Data Modification.md
HeikoTheissen, line 156: Really? A navigation property binding
```xml
<EntitySet Name="Orders">
  <NavigationPropertyBinding Path="Items/CancellationItem" Target="ns.container/Orders/Items" />
</EntitySet>
```
is insufficient to determine the canonical collection for the create request `POST Orders(23)/Items(5)/CancellationItem`. This needs service-specific knowledge that the cancellation item belongs to the same order, hence the collection is `Orders(23)/Items`.

odata-protocol/11.4 Data Modification.md
```diff
@@ -148,6 +148,21 @@ to that collection's URL. The `POST` body MUST contain a single valid
...
+
+or the payload contains a context URL.
+
+The service MUST fail the request if the
+entity represented in the payload exists already in the determined collection.
```
HeikoTheissen, line 161: Or should the POST request then be treated as an update and return 200 OK without Location header?

odata-protocol/11.4 Data Modification.md
ralfhandl: This is not a valid [binding target](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.02/csd01/odata-csdl-xml-v4.02-csd01.html#BindingTarget). 

> If the target is a target path, it MUST resolve to an entity set, singleton, or direct or indirect containment navigation property of a _singleton_ in scope.

With that restriction in mind line 156 should work.

odata-protocol/11.4 Data Modification.md
ralfhandl: This section is about POST to a collection URL. A singleton is _not_ a one-element collection, and OData does not define POST to a singleton URL. We had a long debate on this and decided against it.

So this sentence here does not apply to singletons. 

The referenced sentence on PUT/PATCH to a single entity URL does apply to singletons.

odata-protocol/11.4 Data Modification.md
ralfhandl: No, POST to single entity URL is intentionally undefined.

HeikoTheissen: This is not about POST to a single-entity URL, but to a collection-valued non-containment navigation property, for example
```http
POST Orders(23)/Products

{"ProductID": 5}
```
if Product 5 already exists.

odata-protocol/11.4 Data Modification.md
HeikoTheissen: The term "containment navigation property of a singleton" seems undefined. Is this
- a single-valued containment navigation property?
- a path starting with a singleton that ends in a containment navigation property?

Why are singletons treated specially here?

ralfhandl: > The path can traverse single-valued containment navigation properties or single-valued complex properties before ending in a containment navigation property, and there MUST NOT be any non-containment navigation properties prior to the final segment.

What is missing in this definition?

odata-protocol/11.4 Data Modification.md
ralfhandl: So far such a request always failed, and I would not change that in a minor version.

odata-protocol/11.4 Data Modification.md
HeikoTheissen: This definition is OK, but the additional condition
> If the target is a target path, it MUST resolve to an entity set, singleton, or direct or indirect _containment navigation property of a singleton_ in scope.

leaves unclear where the singleton must be (at the beginning or the end of the path).

ralfhandl: Thanks! Please suggest an addition.

mikepizzo: Discussed 2025-3-26 and agreed that the current wording is sufficient.

docs/odata-url-conventions/odata-url-conventions.md
```diff
@@ -605,8 +605,8 @@ segments to a `resourcePath` that identifies a collection of entities,
 for example by:
 - Using an entity key to select a single
   entity (see rules: `collectionNavigation` and `keyPredicate`).
-  This is only allowed if the service can determine the canonical collection the canonical collection (as defined in [OData-Protocol, section 10](https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part1-protocol.html#ContextURL))
-  from the resource path up to the `keyPredicate`.
+  If the service cannot determine the canonical collection the canonical collection (as defined in [OData-Protocol, section 10](https://docs.oasis-open.org/odata/odata/v4.02/odata-v4.02-part1-protocol.html#ContextURL))
```
uhlmannm: the canonical collection twice?

odata-csdl/13 Entity Container.md
ralfhandl, line 1: Please move lines 437-439 up directly after line 427.

odata-csdl/13 Entity Container.md
```diff
@@ -446,6 +446,10 @@ that differ only in a type-cast segment, allowing to bind instances of
 different sub-types to different targets. If paths differ only in
 type-cast segments, the most specific path applies.
 
+Elsewhere in this specification, an [instance path](#PathExpressions) that starts
```
ralfhandl, line 449: Where? Please add a link.

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
+3. ends with a containment navigation property,
+   followed by a key predicate if this is collection-valued
+   (and which must be nullable otherwise if the request can be treated as an insert)
+4. matches a navigation property path binding [#OData-CSDL#NavigationPropertyPathBinding]
 
-- containment navigation properties,
-- non-containment navigation properties with a navigation property binding, or
-- payloads including a context URL specifying the entity set or
-  contained collection of entities in which the new entity is to be created.
+or the payload includes a context URL.
```
ralfhandl, line 692: Why isn’t this the fifth numbered item in the list?

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
 services MUST either fail the request or ignore the value in the request body.
 
-To ensure that an update request is not treated as an insert, the client
-MAY specify an [`If-Match`](#HeaderIfMatch) header in the update
-request. The service MUST NOT treat an update request containing an
-`If-Match` header as an insert.
+If the URL of an upsert request falls into case 1, 2, or 3 given above and the request
+- contains an [`If-Match: *`](#HeaderIfMatch) header, the service MUST fail the request
+  if it would be treated as an insert
+- contains an `If-Match` header with an ETag as value, the service fails the request
```
ralfhandl, line 712: Why not „the service MUST fail the request“?

odata-url-conventions/4 Resource Path.md
```diff
@@ -145,7 +145,13 @@ Often however a single entity is accessed by composing more path
 segments to a `resourcePath` that identifies a collection of entities,
 for example by:
 - Using an entity key to select a single
-entity (see rules: `collectionNavigation` and `keyPredicate`)
+  entity (see rules: `collectionNavigation` and `keyPredicate`).
+  If the service cannot determine the canonical collection (as defined in [#OData-Protocol#ContextURL])
```
ralfhandl, line 149: This would read nicer if it were a new top-level paragraph after the list.

odata-csdl/13 Entity Container.md
HeikoTheissen: In several places, which all link back here. Shall I also add forward links?

odata-protocol/11.4 Data Modification.md
HeikoTheissen: Items 1-4 have the common prefix "if the resource path of the request", the fifth item is different.

odata-protocol/11.4 Data Modification.md
HeikoTheissen: Because this is not a new requirement, it already follows from section 11.4.1.1.

ralfhandl (dismissed):

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
+Example ##ex: `Category` is a single-valued nullable non-containment navigation property
+on the product entity type. Even if
+```
+GET http://host/service/Products(57)/Category
+```
+returns `404 Not Found`, the upsert request
+```
+PATCH http://host/service/Products(57)/Category
+```
+is treated like an update not an insert if its payload identifies an existing category.
```
HeikoTheissen, line 735: This PATCH request may or may not be treated as an insert, hence the `Upsertable` capability may or may not be relevant. This means that a client cannot know it advance whether it can make this request.

odata-csdl/13 Entity Container.md
```diff
@@ -446,6 +446,11 @@ that differ only in a type-cast segment, allowing to bind instances of
 different sub-types to different targets. If paths differ only in
 type-cast segments, the most specific path applies.
 
+Elsewhere in this specification ([#OData-Protocol#CreateanEntity], [#OData-Protocol#UpsertanEntity],
+[#OData-URL#AddressingEntities]) an [instance path](#PathExpressions) that starts
+with an entity container is said to _match_ a navigation property path binding
```
mikepizzo, line 451: Should be starts at, not starts with (the entity container is not part of the url)

odata-protocol/11.4 Data Modification.md
mikepizzo: I like resource path...

odata-protocol/11.4 Data Modification.md
mikepizzo: This seems to me like it would be contrary to REST patterns.

mikepizzo: 2025-4-16: agreed to keep an error.

odata-protocol/11.4 Data Modification.md
```diff
@@ -314,8 +325,8 @@ On failure, the service MUST NOT create any of the entities.
 
 ### ##subsubsec Update an Entity
 
-To update an individual entity, the client makes a `PATCH` or `PUT`
-request to a URL that identifies the entity. The [edit URL](#ReadURLsandEditURLs)
+To make changes to an entity, the client makes `PATCH` or `PUT`
+request (a so-called _upsert request_) to a URL that identifies the entity. The [edit URL](#ReadURLsandEditURLs)
```
mikepizzo, line 329: I would not use upsert here; depending on whether the if-match it may just be an update.

odata-protocol/11.4 Data Modification.md
```diff
@@ -364,7 +375,10 @@ property `$value`.
 Updating a dependent property that is tied to a key property of the
 principal entity through a referential constraint updates the
 relationship to point to the entity with the specified key value. If
-there is no such entity, the update fails.
+the entity set for that entity cannot be determined or does not contain
```
mikepizzo, line 378: should be canonical collection
```suggestion
the canonical collection for that entity cannot be determined or does not contain
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
-Upserts to single-valued navigation properties are possible for
+If the service is unable to determine the
+canonical collection or canonical singleton (as defined in [section ##ContextURL]) for the newly-created entity,
+it MUST fail the upsert request.
+This determination is possible, for example, if the resource path of the request
+1. consists of an entity set followed by a key predicate
+2. consists of a singleton (which must be nullable if the request can be treated as an insert)
+3. ends with a containment navigation property,
+   followed by a key predicate if this is collection-valued
+   (and which must be nullable otherwise if the request can be treated as an insert)
```
mikepizzo, line 689: I would not try to cover both single- and collection-valued in one bullet; it makes it too confusing (especially with the part about nullability).

mikepizzo: We could combine entity set and collection-value containment navigation property, and combine singleton and single-valued containment navigation property.

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
-To ensure that an update request is not treated as an insert, the client
-MAY specify an [`If-Match`](#HeaderIfMatch) header in the update
-request. The service MUST NOT treat an update request containing an
-`If-Match` header as an insert.
+If the URL of an upsert request falls into case 1, 2, or 3 given above and the request
+- contains an [`If-Match: *`](#HeaderIfMatch) header, the service MUST fail the request
+  if it would be treated as an insert
+- contains an `If-Match` header with an ETag as value, the service fails the request
+  instead of treating it as an insert because the [ETag](#UseofETagsforAvoidingUpdateConflicts)
+  does not match
```
mikepizzo, line 714: this leaves out the fact that the service can treat it as an update.
```suggestion
- contains an `If-Match` header with an ETag as value, the service treats the request
  as an update if the ETag matches the existing entity, otherwise fails the request
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
-
 A key property whose value is provided in the request URL SHOULD be omitted from the request body.
 If key properties are provided in the request URL and the request body with different values,
 services MUST either fail the request or ignore the value in the request body.
 
-To ensure that an update request is not treated as an insert, the client
-MAY specify an [`If-Match`](#HeaderIfMatch) header in the update
-request. The service MUST NOT treat an update request containing an
-`If-Match` header as an insert.
+If the URL of an upsert request falls into case 1, 2, or 3 given above and the request
```
mikepizzo, line 709: would rather not reference numbers here, if we can avoid it.

mikepizzo: maybe specify "the resource path is the canonical url" or something of that sort (if, indeed, we agree that it is different for the canonical url versus a nav prop binding...)

odata-protocol/11.4 Data Modification.md
```diff
@@ -654,49 +668,72 @@ operations.
...
-[`If-None-Match`](#HeaderIfNoneMatch) header is specified with a value
-of `*`.
+In other words, the `If-`(`None-`)`Match` header distinguishes between insert
+and update in these cases.
+
+Note outside of these cases it can happen that an upsert request does not address an existing entity with its URL
+but identifies one in its payload. Such a request may update the identified entity and
+link it to another entity, but it does not lead to a create entity request.
+
+::: example
```
mikepizzo, line 725: consider moving this example (or a discussion of this example) to the etag section.

odata-csdl/13 Entity Container.md
```diff
@@ -446,6 +446,11 @@ that differ only in a type-cast segment, allowing to bind instances of
...
+if the binding applies to all instances that the instance path addresses.
```
mikepizzo, line 452: I'm still a little unclear exactly what we're trying to say here...

odata-protocol/11.4 Data Modification.md
```diff
@@ -148,6 +148,21 @@ to that collection's URL. The `POST` body MUST contain a single valid
...
 or one of its derived types.
 
+If the service is unable to determine the canonical collection (as defined in [section ##ContextURL])
+for the entity, it MUST fail the request.
+This determination is possible, for example, if the collection's URL
+- consists of an entity set
+- ends with a containment navigation property
+- matches a navigation property path binding [#OData-CSDL#NavigationPropertyPathBinding]
+
+or the payload contains a context URL.
```
mikepizzo, line 158: Maybe preface this with something like "The URL specified in the POST request need not be the canonical URL for the collection; for example, it could represent a non-contained navigation property.  In such a case, ..."

And, since we have now scoped it to the non-canonical case, we don't need to talk about 1 & 2, and can just say something like:

```suggestion
The URL specified in the POST request need not be the canonical URL for the collection; for example, it could represent a non-contained navigation property. In such a case, the service can determine the canonical collection through a matching navigation property path binding [#OData-CSDL#NavigationPropertyPathBinding] or the presence of a context URL within the payload. If the service cannot determine the canonical collection, it MUST fail the request.
```

This describes the common cases, but leaves open the option for the service to have other ways to determine the canonical collection.

HeikoTheissen: I like that wording, will use it elsewhere as well.

odata-protocol/11 Data Service Requests.md
```diff
@@ -155,6 +155,18 @@ instance, for example as a `readLink` or `editLink` in an
...
 
+::: example
+Example ##ex: Products can be sourced from a supplier (like `Suppliers(5)`)
+as well as from a subsidiary (like `Subsidiaries(5)`). These two entities have the same
+entity type that is also used by the non-containment navigation property `SourcedFrom`
+defined on the product entity type. Then the following URL identifies a collection of entities followed by
+an entity key to select a single entity (like in [#OData-URL#AddressingEntities])
+but might identify either of the two entities:
+```
+GET http://host/service/Products(1)/SourcedFrom(5)
```
HeikoTheissen, line 166: Since this URL does not identify a unique entity, what shall the response be?

odata-protocol/11.4 Data Modification.md
```diff
@@ -69,10 +69,25 @@ specifying `If-Match` with a value of `*`. Services MAY reject such
...
 
 For requests including an [`OData-Version`](#HeaderODataVersion) header
-value of `4.01`, any ETag values specified in the request body of an
-[update request](#UpdateanEntity) MUST be `*` or match the current value
+value of `4.01`, any ETag values specified in the request body of a
+[request for modifying an entity](#UpdateanEntity) MUST be `*` or match the current value
 for the record being updated.
 
+::: example
+Example ##ex_antietag: `Category` is a single-valued nullable non-containment navigation property
```
mikepizzo, line 77: Note that, (in the given case), the etag used in an if-match or if-not-match header applies to the category currently associated with the product. if you want to make an update conditional on the product, you would instead issue the update against the product, with the updates to the category in the payload.

odata-protocol/11.4 Data Modification.md
```diff
@@ -144,10 +159,24 @@ of these query options and instead MUST return [`204 No Content`](#ResponseCode2
 ### ##subsubsec Create an Entity
 
 To create an entity in a collection, the client sends a `POST` request
-to that collection's URL. The `POST` body MUST contain a single valid
+to a resource path addressing that collection. The `POST` body MUST contain a single valid
 representation of an entity of the declared target entity type,
 or one of its derived types.
 
+The resource path need not be the canonical URL for the collection; for example,
+it could represent a non-contained navigation property.
```
mikepizzo, line 167:
```suggestion
it could represent a non-containment navigation property.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -144,10 +159,24 @@ of these query options and instead MUST return [`204 No Content`](#ResponseCode2
...
+In such a case, the service can determine the canonical collection (as defined in [section ##ContextURL])
+through a navigation property path binding [#OData-CSDL#NavigationPropertyPathBinding]
+that applies to all instances addressed by the resource path
+or through the presence of a context URL within the payload.
+If the service is unable to determine the canonical collection for the entity, it MUST fail the request.
```
mikepizzo, line 172:
```suggestion
In such a case, the service may be able to determine the canonical collection (as defined in [section ##ContextURL]), for example, through a navigation property path binding [#OData-CSDL#NavigationPropertyPathBinding]
or through the presence of a context URL within the payload.
If the service is unable to determine the canonical collection for the entity, it MUST fail the request.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -144,10 +159,24 @@ of these query options and instead MUST return [`204 No Content`](#ResponseCode2
...
+
+The service MUST fail the request if the
+entity represented in the payload exists already in the determined collection.
```
mikepizzo, line 175:
```suggestion
entity represented in the payload already exists in the determined collection.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -314,7 +339,7 @@ On failure, the service MUST NOT create any of the entities.
 
 ### ##subsubsec Update an Entity
 
-To update an individual entity, the client makes a `PATCH` or `PUT`
+To make changes to an entity, the client makes `PATCH` or `PUT`
```
mikepizzo, line 342:
```suggestion
To make changes to an entity, the client makes a `PATCH` or `PUT`
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -364,7 +389,10 @@ property `$value`.
 Updating a dependent property that is tied to a key property of the
 principal entity through a referential constraint updates the
 relationship to point to the entity with the specified key value. If
-there is no such entity, the update fails.
+the canonical collection (as defined in [section ##ContextURL]) for that entity cannot be determined or does not contain
+such an entity, the update fails. The canonical collection can be determined, among
+other possibilities, by containment navigation properties or navigation property bindings
+or by a context URL in the request payload.
```
mikepizzo, line 395:
```suggestion
such an entity, the update fails. The canonical collection is known for containment navigation properties, and can be determined in the presence of navigation property bindings or a context URL in the request payload, or through service specific knowledge.
```

or similar wording as above.

odata-protocol/11.4 Data Modification.md
```diff
@@ -433,7 +461,7 @@ the service MAY ignore the system query options and respond with `204 No Content
 
 #### ##subsubsubsec Update Related Entities When Updating an Entity
 
-Update requests with an OData-Version header with a value of `4.0` MUST
+Requests for changing an entity with an OData-Version header with a value of `4.0` MUST
```
mikepizzo, line 464:
```suggestion
Requests to change an entity with an OData-Version header with a value of `4.0` MUST
```
(throughout)

odata-protocol/11.4 Data Modification.md
```diff
@@ -655,49 +683,51 @@ operations.
...
-
-An upsert occurs when the client sends an [update
-request](#UpdateanEntity) to a valid URL that identifies a single entity
-that does not yet exist. In this case the service MUST handle the
-request as a [create entity request](#CreateanEntity) or fail the
-request altogether.
+#### ##subsubsubsec Upsert an Entity
 
-Upserts to single-valued navigation properties are possible for
+Services MAY treat a request for changing an entity like a [create entity request](#CreateanEntity)
```
mikepizzo, line 688:
```suggestion
Services MAY treat a request to change an entity as a [create entity request](#CreateanEntity)
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -655,49 +683,51 @@ operations.
...
+if the service is unable to determine the canonical collection, it MUST fail the request.
 
 Upserts are not supported against entities whose keys' values are
-generated by the service. Services MUST fail an update request to a URL
+generated by the service. Services MUST fail a request to a URL
 that would identify such an entity and the entity does not yet exist.
 
-Similarly, services MUST fail an update request to the URL of a [media entity](#RequestingtheMediaStreamofaMediaEntityusingvalue) that does not yet exist.
-However, a `PUT` request to the _media edit URL_ of a media entity does have Upsert
+Similarly, services MUST fail a `PATCH` request to the URL of a [media entity](#RequestingtheMediaStreamofaMediaEntityusingvalue) that does not yet exist.
```
mikepizzo, line 707:
```suggestion
Similarly, services MUST fail an update request to the URL of a [media entity](#RequestingtheMediaStreamofaMediaEntityusingvalue) that does not yet exist.
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -655,49 +686,50 @@ operations.
...
+- contains an `If-Match` header with an ETag as value, the service treats the request as
+  an update if the [ETag](#UseofETagsforAvoidingUpdateConflicts) matches the addressed entity
+  or fails the request, but never treats it as an insert
+- contains an [`If-None-Match: *`](#HeaderIfNoneMatch) header, the service MUST fail the request
+  if it would be treated as an update.
+
+In other words, the `If-`(`None-`)`Match` header distinguishes between insert
+and update in this case. Otherwise the resource path ends with a
+non-containment navigation property and an `If-`(`None-`)`Match` header applies to
+the current related entity addressed by it (see [example ##antietag]).
```
HeikoTheissen, line 732: In the other case, the client may not care whether the operation is an update or upsert, and hence may not want to evaluate both `Updatable` and `Upsertable` in the capabilities annotations.

odata-protocol/11 Data Service Requests.md
```diff
@@ -155,6 +155,24 @@ instance, for example as a `readLink` or `editLink` in an
 MAY support conventions for constructing a read URL using the entity's
 key value(s), as described in [#OData-URL#CanonicalURL].
 
+Borderline cases are possible in which two or more entity sets with the same entity type
+use overlapping keys and a non-containment navigation property [#OData-CSDL#ContainmentNavigationProperty]
+with that entity type does not have a unique navigation property binding [#OData-CSDL#NavigationPropertyBinding].
+In such cases, a URL that identifies a collection of entities followed by
+an entity key to select a single entity (like in [#OData-URL#AddressingEntities])
+may not identity a unique entity, and the behavior of the service is undefined.
```
mikepizzo, line 163:
```suggestion
may not identify a unique entity, and the behavior of the service is undefined.
```

mikepizzo (dismissed):

mikepizzo (approved):

PR #2064, opened 2025-03-20
ETags in update responses

### LINKED ISSUE #2040: ETags in responses with `return=minimal`
[OData-Protocol, section 11.4.3](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#UpdateanEntity) states
> If the service uses ETags for optimistic concurrency control, the entities in the response MUST include ETags.

Does this also apply if the `Preference: return=minimal` is applied? The OData [specification](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#Preferencereturnrepresentationandreturnminimal) of this preference (and even more the underlying [RFC 7240](https://www.rfc-editor.org/rfc/rfc7240.html#section-4.2)) seems to leave open the possiblity of a response with
```
200 OK
Preference-Applied: return
```
and only keys and etags in the payload.

Clarify this in the OData specification.

HeikoTheissen:
Fixes #2040


odata-protocol/11.4 Data Modification.md
```diff
@@ -425,9 +425,10 @@ request that the response SHOULD include a body by specifying a
 [`return=representation`](#Preferencereturnrepresentationandreturnminimal) preference, or by
 specifying the system query options
 [`$select`](#SystemQueryOptionselect) or
-[`$expand`](#SystemQueryOptionexpand). If the service uses ETags for
-optimistic concurrency control, the entities in the response MUST
-include ETags. If a representation of the updated entity could not be constructed,
+[`$expand`](#SystemQueryOptionexpand).
+Any entity created or updated as part of the request that requires concurrency control
```
ralfhandl, line 429:
```suggestion
Any entity that requires concurrency control
```
would be easier to implement, and easier to understand for the client. It might be tricky to determine whether an entity was "created or updated as part of the request".

ralfhandl (approved):

PR #2073, opened 2025-04-01
Clarify that navigation properties with no keys defined require a cast segment

### LINKED ISSUE #361: Define semantics for navigation properties of type Edm.EntityType
[OData Common Schema Definition Language (CSDL) XML Representation Version 4.01 (oasis-open.org)](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_BuiltInAbstractTypes) defines the behavior for Edm.EntityType. It specifically disallows use of Edm.EntityType as the type of an EntitySet, because all instances within an entity set must have the same key.  However, it leaves open the ability to use Edm.EntityType as the entity type of a navigation property. And, indeed, [ODATA-1552](https://issues.oasis-open.org/browse/ODATA-1552) updates the xsd to allow navigation properties of type Edm.EntityType.

What is missing is a description of the semantics of a navigation property whose type is Edm.EntityType. Specifically, since Edm.EntityType doesn't define a key, it's not possible to index into a navigation property of type Edm.EntityType by key *unless you insert a cast segment*.

This is ramification is implied by other rules (so calling it out is not a breaking change) but it should be explicitly called out so that the user doesn't have to derive the semantics from the existing rules.

We do separately have a rule that requires a collection-valued containment navigation property define a key (and thus, Edm.EntityType cannot be used as the type of a collection-valued containment navigation property):

"For a collection-valued containment navigation property the specified entity type MUST have a [key](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Key) defined."

This is consistent with entitysets, which disallow the use of Edm.EntityType "because all entities in an entity set must have the same key fields to uniquely identify them within the set".

Are there any other semantics of a (collection-valued) navigation property of Edm.EntityType that need to be spelled out?

Also note that, in general, if there is no navigation property binding for a non-containment collection-valued navigation property, the key may not uniquely identify the instance. This has nothing to do with Edm.EntityType, and we may want to decide to open a separate issue to track, it just came up in scope of the discussion. 

### Proposal

Clarify that, if the entity type defined for a collection-valued navigation property does not define a key (including Edm.EntityType), indexing the collection by key requires a cast segment to an entity type with a key defined, and will return the entity of that type (or subtype) with the specified key, if such an instance exists.

Imported from [ODATA-1558](https://issues.oasis-open.org/browse/ODATA-1558)

mikepizzo:
Fixes #361


odata-url-conventions/4 Resource Path.md
```diff
@@ -373,7 +382,9 @@ http://host/service/Employees('A1245')
 
 Services MAY support an alternate convention for addressing entities by
 appending a segment containing the unprefixed and unquoted key value to the URL of the
-collection containing the entity. Forward-slashes in key value segments
+collection containing the entity. If the declared entity type of the collection does not 
+define a key (for example, Edm.EntityType), then a cast segment to a type defining a key 
+MUST follow the collection, before appending the key value segment. Forward-slashes in key value segments
```
ralfhandl, line 387: This is a potentially breaking change, existing services could in 4.01 allow key-as-segment without a type-cast segment.

mikepizzo: A type-cast segment should be valid for key-as-segment as well.

HeikoTheissen (approved):

PR #2076, opened 2025-04-10
Upsert actions

### LINKED ISSUE #2071: Can actions behave like upsert operations?
[OData-Protocol, section 11.5.5.1](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#InvokinganAction) states
> Actions that create and return a single entity follow the rules for [entity creation](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#CreateanEntity) and return a [Location](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#HeaderLocation) header that contains the edit URL or read URL of the created entity.

That seems to imply that the "create-like behavior" is a property of a the action as a whole, so that an action cannot
- sometimes create an entity (return "201 Created" and a Location header) and
- at other times update an entity (return "200 OK").

In other words: An action cannot behave like an upsert.

Can we replace
> Actions that create and return a single entity follow ... and return ...

with
> If an action creates and returns a single entity, it follows ... and returns ...

?

HeikoTheissen:
Fixes #2071


odata-protocol/11.5 Operations.md
```diff
@@ -423,10 +423,10 @@ default content type will be used.
 The client can request whether any results from the action be returned
 using the [`return`](#Preferencereturnrepresentationandreturnminimal) preference.
 
-Actions that create and return a single entity follow the rules for
-[entity creation](#CreateanEntity) and return a
+If an invocation of an action results in the creation and return of a single entity,
```
ralfhandl, line 426: This reads better, but I can't quote a grammar or style rule justifying that change:

```suggestion
If the invocation of an action results in the creation and return of a single entity,
```

ralfhandl (dismissed):

ralfhandl (approved):

mikepizzo (approved):

PR #2081, opened 2025-05-02
More explicit definition of EntitySetPath

### LINKED ISSUE #2065: Exact meaning of EntitySetPath: same order or some order?
Consider a bound operation with entity set path:
```xml
<Action Name="Act" IsBound="true" EntitySetPath="Order/Items">
  <Parameter Name="Order" Type="self.Order" />
  <ReturnType Type="self.OrderItem" />
</Action>
```
where `Items` is a containment navigation property.

Does this mean that the return value is contained in `Orders('A')/Items` for _some_  key `A`? Or for the key of the binding parameter (`Order/ID = 'A'`)?

In the first case, the `EntitySetPath` would not define an entity set. It would be analogous to a navigation property binding with target path `ns.Container/Orders/Items`, which is forbidden.

HeikoTheissen:
Fixes #2065


odata-csdl/12 Action and Function.md
```diff
@@ -181,22 +181,49 @@ Absence of the attribute means `false`.
 
 ## ##subsec Entity Set Path
 
-Bound actions and functions that return an entity or a collection of
-entities MAY specify an entity set path if the entity set of the
-returned entities depends on the entity set of the binding parameter
+Actions and functions bound to an entity or a collection of
+entities that return an entity or a collection of
```
HeikoTheissen, line 185: If $k>0$, the binding parameter need not be an entity (collection), it can also be a complex instance.

Example: The return type is the entity type `Country`, the binding parameter type is the complex type `Address` with a containment navigation property `Country`. The entity set path is `bindingParameter/Country`.

When the operation is invoked on `Customers(2)/Address`, the returned entities MUST belong to the implicit entity set `Customers(2)/Address/Country`.

odata-csdl/12 Action and Function.md
```diff
@@ -186,6 +186,9 @@ entities MAY specify an entity set path. The entity set path specifies the canon
 (as defined in [#OData-Protocol#ContextURL]) of the
 returned entities in terms of the binding parameter value.
 
+If no entity set path is specified or the canonical collection cannot be determined
+by the following rules, the service SHOULD include a context URL [#OData-Protocol#ContextURL] in the response.
+
```
HeikoTheissen, line 191: Delete this paragraph and instead rewrite [OData-Protocol, sections 10.2 and 10.3] where the entity set path is mentioned.

mikepizzo (approved):

PR #2082, opened 2025-05-12
Update -> upsert

### LINKED ISSUE #2060: Nullable single-valued navigation property or singleton: update or upsert?

HeikoTheissen:
Fixes #2060


ralfhandl (commented):
+1, only minor remarks

odata-protocol/11.4 Data Modification.md
```diff
@@ -701,7 +701,7 @@ through a navigation property path binding [#OData-CSDL#NavigationPropertyPathBi
 or through the presence of a context URL within the payload;
 if the service is unable to determine it, it MUST fail the request.
 
-Upserts are not supported against entities whose keys' values are
+Upsert requests MUST NOT be treated as inserts of entities whose keys' values are
```
mikepizzo, line 704: it's really can not -- there is no way to specify the url of an entity whose key is generated by the service and does not yet exist.  The MUST comes in the following line.

ralfhandl (approved):

PR #2087, opened 2025-06-03
Avoid flickering when navigating from dark mode to dark mode

HeikoTheissen:


ralfhandl (approved):
Trusting that you have tested this 😁

PR #2129, opened 2025-10-02
URI fragments in @type control information

### LINKED ISSUE #2127: Position of # in @type references
[OData-JSON, section 4.6.3](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#ControlInformationtypeodatatype) states
> For non-built in primitive types, the URI contains the namespace-qualified or alias-qualified type, specified as a URI fragment.

But if the type is defined externally, as in [this example](https://github.com/oasis-tcs/odata-vocabularies/blob/c8176abe3e18cf238e183e7aae6ee120948482d6/examples/Org.OData.Temporal.V1.objectkey-sample.json#L55)
```json
"@odata.type": "https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Temporal.V1.xml#Temporal.UnitOfTimeDate"
```
this is an entire URI with fragment, not just a URI fragment.

> For properties that represent a collection of values, the fragment is the namespace-qualified or alias-qualified element type enclosed in parentheses and prefixed with Collection.

"the fragment is the ... type" sounds as if there would be no `#` character. And it leaves unclear whether the format is `Collection(#ns.type)` or `#Collection(ns.type)`.

Only the former works in the case of externally defined types compare `Collection(https://path/to/csdl#ns.type)` with the non-sensical `#Collection(https://path/to/csdlns.type)`.

HeikoTheissen:
Fixes #2127


odata-json-format/4 Common Characteristics.md
```diff
@@ -281,10 +281,10 @@ using the `$schemaversion` system query option
 defined in [#OData-Protocol#SystemQueryOptionschemaversion].
 
 For non-built in primitive types, the URI contains the
-namespace-qualified or alias-qualified type, specified as a URI
-fragment. For properties that represent a collection of values, the
-fragment is the namespace-qualified or alias-qualified element type
-enclosed in parentheses and prefixed with `Collection`. The
+namespace-qualified or alias-qualified type, specified in the URI
+fragment (after the `#`). For properties that represent a collection of values, the
```
ralfhandl, line 285: I stumble over the "in the" and would prefer something like

```suggestion
namespace-qualified or alias-qualified type, specified as a URI
fragment (`#` followed by the namespace-qualified or alias-qualified element type name). For properties that represent a collection of values, the
```

PR #2137, opened 2025-10-23
`content-type` header in JSON batch

### LINKED ISSUE #2134: `content-type` header in JSON batch requests
https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#BatchRequest states
> The value of `headers` is an object whose name/value pairs represent request headers. The name of each pair MUST be the lower-case header name; the value is a string containing the header-encoded value of the header. The `headers` object MUST contain a name/value pair with the name `content-type` whose value is the media type.

This is wrong, because a `content-type` header is mandatory only if a `body` is present, but not for `GET` or `DELETE` requests. However, in [example 55](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#batchRequest)
* the `content-type` is required but missing, also
* the `Prefer` header name is upper-case.

HeikoTheissen:
Fixes #2134


odata-json-format/19 Batch Requests and Responses.md
```diff
@@ -383,11 +393,9 @@ value is a string containing the header-encoded value of the header.
...
 The response object MAY contain the name/value pair `body`
 which follows the same rules as within [request objects](#BatchRequest).
-
-If the media type is not exactly equal to `application/json`
-(i.e. it is a subtype or has format parameters), the
-`headers` object MUST contain a name/value pair with the name
-`content-type` whose value is the media type.
+The `content-type` header mandated by [#OData-Protocol#HeaderContentType]
+MAY be omitted if the media type is exactly equal to `application/json`
+(i.e. it is not a subtype and has no format parameters).
```
HeikoTheissen: Is the same exception also made for the `content-type` header in JSON batch _requests_?

odata-json-format/19 Batch Requests and Responses.md
```diff
@@ -384,10 +387,9 @@ value is a string containing the header-encoded value of the header.
 The response object MAY contain the name/value pair `body`
 which follows the same rules as within [request objects](#BatchRequest).
 
-If the media type is not exactly equal to `application/json`
-(i.e. it is a subtype or has format parameters), the
-`headers` object MUST contain a name/value pair with the name
-`content-type` whose value is the media type.
+The `content-type` header mandated by [#OData-Protocol#HeaderContentType]
+MAY be omitted if the media type is exactly equal to `application/json`
+(i.e. it is not a subtype and has no format parameters).
```
mikepizzo:
```suggestion
If the header property of the response object does not define the content-type, then the content-type header mandated by [#OData-Protocol#HeaderContentType] is assumed to be `application/json` (without any format parameters).
```

odata-json-format/19 Batch Requests and Responses.md
```diff
@@ -119,6 +118,10 @@ containing the base64url-encoded value of the request body. In this case
 the body content can be compressed or chunked if this is correctly
 reflected in the `Transfer-Encoding` header.
 
+The `content-type` header mandated by [#OData-Protocol#HeaderContentType]
+MAY be omitted if the media type is exactly equal to `application/json`
+(i.e. it is not a subtype and has no format parameters).
+
```
mikepizzo:
```suggestion
Services MAY support omitting the content-type in the header property of a request object. Such requests MUST be interpreted as if the `content-type` header mandated by [#OData-Protocol#HeaderContentType] were specified with a value of `application/json` (with no format parameters).
```

odata-json-format/19 Batch Requests and Responses.md
```diff
@@ -101,8 +101,7 @@ atomicity group, all requests in that group fail with
 The value of `headers` is an object whose name/value pairs
 represent request headers. The name of each pair MUST be the lower-case
 header name; the value is a string containing the header-encoded value
-of the header. The `headers` object MUST contain a name/value
-pair with the name `content-type` whose value is the media type.
```
mikepizzo: This should be restored (at least as guidance to the client) and co-located with the new paragraph describing the behavior of services that allow omitting the content-type.

PR #2158, opened 2026-01-14
Define Composability and use of query options for actions/functions (Fixes OData-400)

### LINKED ISSUE #400: Access a property of the result of a non-composable function
Must a function be composable to allow access to individual properties of its returned value?
```java
GET BillingDocument('90000090')/self.GetPDF()/BillingDocOutputDataBinary
```
The [CSDL specification](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_ComposableFunction) says
> A composable function can be invoked with additional path segments or key predicates appended to the resource path that identifies the composable function, and with system query options as appropriate for the type returned by the composable function.
> 

and `BillingDocOutputDataBinary` is an "additional path segment".

The [Protocol specification](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_InvokingaFunction) seems to allow system query options without composability
> 
> If the function is composable, additional path segments may be appended to the URL that identifies the composable function (or function import) as appropriate for the type returned by the function (or function import). The last path segment determines the system query options and HTTP verbs that can be used with this this URL
> 

Assuming that `BillingDocOutputDataBinary` is a stream property,
```java
GET BillingDocument('90000090')/self.GetPDF()?$expand=BillingDocOutputDataBinary
```
is allowed without composability, and it requires the server to retrieve the same data (just serialize it differently).

 Related to [ODATA-1644](https://issues.oasis-open.org/browse/ODATA-1644)

### Proposal

Step 1: come up with one definition of non/composability, also taking actions into account which are "non-composable" now

Step 2: complement it with capabilities to express what actually is possible if a function is "composable"

Imported from [ODATA-1640](https://issues.oasis-open.org/browse/ODATA-1640)

mikepizzo:
Fixes #400.
-Composable Functions support appended segments.
-Capabilities Vocabulary can specify supported query options for compoosable functions, non-composable functions, and actions.

Questions:
 - [ ] Should composable functions support query options by default, while non-composable and actions need to specify, or should all actions/functions be annotated for any query options they support?
 - [ ] Should only composable functions be indexable by key?
 - [ ] Verify References to Capabilities vocabularies


HeikoTheissen (dismissed):

odata-csdl/12 Action and Function.md
```diff
@@ -12,7 +12,9 @@ any type.
 The action's name is a [simple identifier](#SimpleIdentifier) that MUST
 be unique within its schema.
 
-Actions cannot be composed with additional path segments.
+Actions cannot be composed with additional path segments and SHOULD 
+use appropriate Capabilities vocabulary annotations [OData-VocCap](#ODataVocCap)
+to denote any supported query options.
```
HeikoTheissen: Do you mean the capabilities annotations that target the entities returned by the action (independent of the action)? Or capabilities annotations that have the action in their annotation target?

Does the phrase "SHOULD use ... to denote" imply that the defaults explained in the capabilities vocabulary do not apply? 

For example, in the absence of capabilities annotations, `$filter` is supposed to be supported. Does this apply here?

mikepizzo: "Functions SHOULD be annotated with the [Capabilities Vocabulary](ODataVocCap) 
annotations appropriate for the return type to denote any supported query options.
Absent such annotations, composable functions SHOULD support the same default query options 
as an entity set of that type would support, while non-composable functions SHOULD NOT
be assumed to support any query options."

"Actions cannot be composed with additional path segments and SHOULD
be annotated with the [Capabilities Vocabulary](ODataVocCap) annotations 
appropriate for the return type to denote any supported query options.
Absent such annotations, actions SHOULD NOT be assumed to support any query options."

HeikoTheissen (approved):

HeikoTheissen (approved):

PR #2164, opened 2026-01-22
Atomicity group of prerequisite requests optional in `dependsOn`

### LINKED ISSUE #2150: A request `dependsOn` another from a different atomicity group
[OData-JSON, section 19.1](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#BatchRequest) states
> If the first segment of a relative path starts with a $ character ..., then this first segment is replaced with the URL of the entity created by or returned from a preceding request whose id value is identical to the value of the first segment with the leading $ character removed. The id of this request MUST be specified in the `dependsOn` name/value pair.

and later
> If a request depends on another request that is part of a different atomicity group, the atomicity group MUST be listed in `dependsOn`.

Does that mean that a request that references another from a different atomicity group must include both IDs (of the referenced request and its atomicity group) in its `dependsOn`?

HeikoTheissen:
Fixes #2150


mikepizzo:
Proposed wording tweak in dependent PR https://github.com/oasis-tcs/odata-specs/pull/2173..

PR #2171, opened 2026-02-12
Encourage chunk-by-chunk processing

### LINKED ISSUE #2163: Clarification requested: JSON $batch response status 200 OK even for malformed JSON requests?
[OData-Protocol, section 11.7](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#BatchRequests) states
> If the set of request headers of a batch request are valid the service MUST return a [200 OK](https://oasis-tcs.github.io/odata-specs/odata-protocol/odata-protocol.html#ResponseCode200OK) HTTP response code to indicate that the batch request was accepted for processing, even if the processing is yet to be completed.

Does this apply even if the request body isn't well-formed JSON at all? In such a case, the body parser may already respond with 400 Bad Request before the OData service implementation gets to see the request.

Or if the request does not meet the requirements of [OData-JSON, section 19.1](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#BatchRequest)
> A JSON batch request body consists of a single JSON object that MUST contain the name/value pair requests and MAY contain [annotations](https://oasis-tcs.github.io/odata-specs/odata-json-format/odata-json-format.html#InstanceAnnotations). It does not contain the context control information.

HeikoTheissen:
Fixes #2163

### DIFF
--- a/docs/odata-protocol/odata-protocol.html
+++ b/docs/odata-protocol/odata-protocol.html
@@ -1286,8 +1286,7 @@ This uses pandoc 3.8.3 from https://github.com/jgm/pandoc/releases/tag/3.8.3.
 <details open><summary>
 <h2 id="95-in-stream-errors"><a id="InStreamErrors" href="#InStreamErrors">9.5 In-Stream Errors</a></h2>
 </summary>
-<p>Because the HTTP response status code is sent before the body of a response, services may encounter an error in generating the response body after having already returned a success status.</p>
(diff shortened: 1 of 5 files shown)


ralfhandl (approved):

mikepizzo:
1) avoid "chunk"
2) maybe turn around: something like: "Services may start returning results before the full response is generated. In this case, the service may send a 200 OK response and later encounter an error. In this case, the service returns an in-stream error, indicating that an error occurred while processing the request generating the result."

PR #2180, opened 2026-03-10
CSDL: mention Core.Subsets

### LINKED ISSUE #366: Remove restriction on Partner navigation
Consider the following scenario:
- Foo
  - allBar Type=Collection(Bar) Partner=foo
  - redBar Type=Collection(Bar) Partner=foo
  - greenBar Type=Collection(Bar) Partner=foo
- Bar
  - foo Type=Foo Partner=allBar

These partner navigation properties satisfy the basic rules
> If the partner navigation property is single-valued, it MUST lead back to the source entity from all related entities. If the partner navigation property is collection-valued, the source entity MUST be part of that collection.
> 

In addition this model clearly expresses that the pair (Foo/allBar, Bar/foo) forms a bi-directional relationship, and the other two navigation properties on Foo are uni-directional with a back-link.

Unfortunately this model violates the last paragraph in CSDL section [8.3 Partner Navigation Property](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_PartnerNavigationProperty)
> If a partner navigation property is specified, this partner navigation property MUST either specify the current navigation property as its partner to define a bi-directional relationship or it MUST NOT specify a partner navigation property. The latter can occur if the partner navigation property is defined on a complex type, or if the current navigation property is defined on a type derived from the type of the partner navigation property.
> 

A spec-compliant model would have to hide useful information, either
- remove Partner=allBar from Bar/foo, or
- remove Partner=foo from Foo/redBar and Foo/greenBar
  

### Proposal

Replace the requirement with an explanation

> 
> If a partner navigation property is specified, and this partner navigation property specifies the current navigation property as its partner, the two navigation properties define a bi-directional relationship. A partner navigation property may specify a navigation property different from the current navigation property. 
> 

 

Imported from [ODATA-1575](https://issues.oasis-open.org/browse/ODATA-1575)

ralfhandl:
Part of
* https://github.com/oasis-tcs/odata-specs/issues/366


chrisspre (dismissed):

Hubert-Heijkers (commented):
Didn't we agree we would consider a different name for this annotation so that we could apply it to single valued navigation properties as well to indicate that the related entity would be come from the targeted collection?

mikepizzo (approved):

PR #2199, opened 2026-04-08
Ordering of Properties in JSON Batch Request/Response Objects.

### LINKED ISSUE #351: OData 4.01 JSON batch format doesn't support streaming use cases well
JSON Format 19.5 Batch Response states:

"In a response to a batch request using the JSON batch request format specified in the preceding section the response objects MAY appear in any order, and each response object MUST contain the id name/value pair with the same value as in the corresponding request object."

This presents difficulties for clients that may wish to stream the processing of responses.

For example, a client sends a batch of GET requests, each of which is expected to return a large volume of data without having all of it in memory at once.

It is reasonable to expect that in order to correctly process a response object, the client needs to at least get the "id" from the response object and then it will be able to determine which of its requests the response object is associated with. However since payload ordering constraints in section 4.4 don't require the "body" of a batch response object to appear after the "id" of the batch response object, the client may need to buffer the entire response object in memory before being able to process it, which entirely defeats the client's intention of streamed batch response processing.

Note: section 4.4 does mention "id" control information in relation to payload ordering, but the "id" in a batch response object is not "id" control information, it is rather the equivalent of a Content-ID.

For multipart/mixed batch responses, effective response streaming can already be achieved.

### Proposal

Define a mechanism by which the client can expect "id" to appear before "body" in batch response objects.

Imported from [ODATA-1534](https://issues.oasis-open.org/browse/ODATA-1534)

mikepizzo:
Fixes #351

### DIFF
--- a/odata-json-format/1 Introduction.md
+++ b/odata-json-format/1 Introduction.md
@@ -48,6 +48,10 @@ Allow common expressions in action payloads|
 [Section ##BatchRequestsandResponses]| 
 Atomicity group of prerequisite requests optional in `dependsOn`| 
 [2150](https://github.com/oasis-tcs/odata-specs/issues/2150)
+Ordering of properties in JSON Batch Request/Response objects| 
+[351](https://github.com/oasis-tcs/odata-specs/issues/351)
+Clients and services MAY require ordered JSON payloads| 
+[351](https://github.com/oasis-tcs/odata-specs/issues/351)
 
 ## ##subsec Glossary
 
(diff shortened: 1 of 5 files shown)


ralfhandl (dismissed):

odata-json-format/1 Introduction.md
```diff
@@ -48,6 +48,8 @@ Allow common expressions in action payloads|
 [Section ##BatchRequestsandResponses]| 
 Atomicity group of prerequisite requests optional in `dependsOn`| 
 [2150](https://github.com/oasis-tcs/odata-specs/issues/2150)
+Ordering of properties in JSON Batch Request/Response objects|
```
HeikoTheissen, line 52:
```suggestion
[Section ##BatchRequestsandResponses]| 
Ordering of properties in JSON Batch Request/Response objects| 
```

HeikoTheissen (approved):

PR #2200, opened 2026-04-08
Add reference to jsonstreamingsupport capability term

mikepizzo:
Adds reference to vocabulary term


mikepizzo:
We decided not to add this term at this time as requiring existing clients to comply with ordering requirements for json request payloads would be a breaking change.

PR #2223, opened 2026-05-26
Add new `delta` format parameter to return=representation

### LINKED ISSUE #309: PATCH: new preference return=delta to get back a delta response
Update requests can have side-effects, e.g. changing the street of an address could as a side-effect also change the zip code even if it is not present in the PATCH request body.

We need a way to communicate to the client what actually has changed.

### Proposal

Add preference `return=delta` for PATCH requests. The server then may respond with a delta payload containing the changes to the server state identified by the ETag in the PATCH request.

It is a preference, so servers can ignore it.

The maximum shape of the delta can be specified via $select and $expand.

Add boolean property `DeltaResponseSupported` to type `Capabilities.UpdateRestrictionsType` to advertise support for this feature.

Imported from [ODATA-1340](https://issues.oasis-open.org/browse/ODATA-1340)

mikepizzo:
Fixes #309.

Use of format parameter for patch requests request that the response be returned in delta format.
Format parameter is ignored where not applicable.


odata-protocol/11.4 Data Modification.md
```diff
@@ -755,11 +755,14 @@ the request.
...
-In the absence of [`$expand`](#SystemQueryOptionexpand), an update request that includes related entities SHOULD include those related entities in the response.
+In the absence of [`$expand`](#SystemQueryOptionexpand), an update request
+that includes related entities SHOULD include those related entities in the response.
 
 If a collection representing the full set of related entities is included in
 the update request, then the full set of related entities for that collection
-SHOULD be included in the response.
+SHOULD be included by default in the response. The client MAY request that related entities
+in such a request be returned as a delta collection containing at least the applied changes
+by specifying the preference [`return=representation;delta`](#Preferencereturnrepresentationandreturnminimal).
```
HeikoTheissen: Since this restricts the effect of the delta preference to related entities, consider the following wording:
```suggestion
SHOULD be included by default in the response, and a preference for delta response [`return=representation;delta`](#Preferencereturnrepresentationandreturnminimal) applied only to related entities.
```
(Otherwise it sounds like a contradiction to section 8.)

mikepizzo: Intention was that specifying the preference is an exception to the initial statement (represent as full).

Recommend rewording: Unless the return=representation;delta preference is applied....

mikepizzo: Also need to update the abnf in the 4.02 branch...

mikepizzo (commented):
changes from discussion.

HeikoTheissen (approved):

PR #2225, opened 2026-05-28
Idempotency of delete requests

### LINKED ISSUE #2103: Interdependent DELETE requests in a change set
Consider a batch request with a change set/atomicity group that contains DELETE requests for two entities:
```
DELETE OrgUnits(1)
DELETE OrgUnits(2)
```
If `DELETE OrgUnits(1)` implicitly deletes `OrgUnits(2)` (because it is a subordinate org unit), what is the response code of `DELETE OrgUnits(2)`?
- Is it 404 Not Found because the entity does not exist?
- Or is it 204 No Content because the change set was successfully executed overall?

Analogous question for
```http
PATCH /OrgUnits

{"@context": "#$delta",
  "value": [{
    "@removed": {"reason": "deleted"},
    "ID": 1
  }, {
    "@removed": {"reason": "deleted"},
    "ID": 2
  }]}
```
Is there a `Core.DataModificationException` for the second deletion?

HeikoTheissen:
Fixes #2103


odata-protocol/11.4 Data Modification.md
```diff
@@ -1254,7 +1258,8 @@ contain any system query options that affect the shape of the result.
 
 Added/changed entities are applied as [upserts](#UpsertanEntity), and
 deleted entities as [deletions](#DeleteanEntity). Non-key properties of
-deleted entities are ignored. The top-level collection may include added
+deleted entities are ignored, deletions of non-existing entities MUST be
+treated as success (idempotency). The top-level collection may include added
```
mikepizzo: Suggestions:
1. Start a new stentence
2. Clarify that entity must be "valid" (although definition of valid is up to service). i.e., if someone specifies an invalid key value (an odd number, where the service only supports even) the service may return an error.
3. Don't try to justify why here (we don't explain, for example, why added/changed are treated as upsert). If we want to explain that applying delta updates are (to the extent possible) idempotent, we should make that statement elsewhere.

```suggestion
deleted entities are ignored. Deletions of valid, non-existing (eg, deleted) entities
MUST be treated as success. The top-level collection may include added
```

odata-protocol/11.4 Data Modification.md
```diff
@@ -1074,7 +1077,8 @@ defined in [OData-VocCap](#ODataVocCap).
 
 A successful `DELETE` request to the edit URL of a stream property
 attempts to set the property to null and results in an error if the
-property is non-nullable.
+property is non-nullable. Services MAY treat deletion of
+a non-existing stream value as success, thus making the `DELETE` request idempotent.
```
mikepizzo: this one is not required (just saying that setting a null stream to null is allowed).

```suggestion
property is non-nullable.
```

mikepizzo (approved):

Issue #2230, opened 2026-06-06
Define conformance requirements for 4.02


mikepizzo:
#2222 Defines the 4.02 version of the protocol with no new conformance requirements.

Should there be any additional conformance requirements for minimal, intermediate, or full 4.02 services?
Should there be any additional conformance requirements for interoperable 4.02 clients?

We should go through new 4.02 features and see which conformance level (if any) they should live in.
Should we require interoperable 4.02 clients and advanced compliant services to always use ordered json?

mikepizzo:
# 4.02 CSDL Features
Notes:
 1) There don't appear to be any feature/conformance differences between xml and json
 2) Current conformance section starts out saying what a 4.01 service should not do in a 4.0 response, but then says what a 4.01 (and 4.02) service shouldn't do that was defined in earlier version(s).

CSDL Changes in 4.02:
1. Allow stream-valued non-binding parameters (#525)
_Propose_: Add to new section "Max version 4.01 or less": Don't return functions containing stream-valued non-binding parameters to 4.01 or less

2. SRID value variable is deprecated (#1935)
-_Propose_: Add that 4.02 should not use `variable` SRID

3. Actions and functions can take, and return, delta payloads (#348)
-_Propose_: Don't return actions/functions that take/return delta payloads to 4.01 or earlier clients?

4. Returned collections of entities may contain null values (#1983)
We previously said that services should not specify nullable for collection-valued return types of entities. Might
this break a 4.01 client not expecting nullable for returned entity collections?
-_Propose_: Don't return actions/functions that return entity collections that may contain nulls

5. Constant Geo values in annotations (#654)
-Already in CSDL Conformance "SHOULD NOT include constant Geo...in annotations" =>clarify this is should not in a 4.0/4.01 payload

6. Constant Stream values in annotations (#654)
-Already in CSDL Conformance: "SHOULD NOT include ... Stream values in annotations" =>clarify this is should not in a 4.0/4.01 payload

7. New path evaluation rules for annotations targeting annotations and external targeting via container (#575)
-we could say that 4.02 (intermediate?) must use the new external targeting. we could say that services should avoid container-bound targets with paths, because they were ambiguous in 4.0/4.01 (but clients should just silently ignore).  Note that, in this specification, we should not give a preference to one style of capability annotation over another, since this document should only be describing how to use/interpret the syntax of the CSDL.

8. Nested If without else part in collections (#326)
[...]

PR #2239, opened 2026-07-09
Use of QUERY verb

### LINKED ISSUE #355: Section 4.17: support new HTTP QUERY method, deprecate /$query path suffix
The IETF Draft [The HTTP QUERY Method](https://httpwg.org/http-extensions/draft-ietf-httpbis-safe-method-w-body.html) defines a new HTTP method `QUERY` that basically is a "GET with request body", being 
- safe
- idempotent
- cacheable, with the request body incorporated in the cache key

This could replace the OData-specific solution `POST .../$query` defined in section [4.17 Passing Query Options in the Request Body](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#sec_PassingQueryOptionsintheRequestBody).

### Proposal

Rephrase section 4.17 to use the new `QUERY` method with a request body of `Content-Type: application/x-www-form-urlencoded` or `application/json` as defined by [ODATA-1405](https://issues.oasis-open.org/browse/ODATA-1405).

Require all query options to be passed in the request body (rationale: mixed locations for query options are an unnecessary complication). 

Deprecate the current approach using `POST` and path suffix `/$query`.

Imported from [ODATA-1544](https://issues.oasis-open.org/browse/ODATA-1544)

ralfhandl:
with original URL as alternative to POST with suffix /$query

Fixes 
* #355

### DIFF
--- a/docs/odata-url-conventions/odata-url-conventions.html
+++ b/docs/odata-url-conventions/odata-url-conventions.html
@@ -1033,13 +1033,22 @@ This uses pandoc 3.8.3 from https://github.com/jgm/pandoc/releases/tag/3.8.3.
 <h2 id="417-passing-query-options-in-the-request-body"><a id="PassingQueryOptionsintheRequestBody" href="#PassingQueryOptionsintheRequestBody">4.17 Passing Query Options in the Request Body</a></h2>
 </summary>
 <p>The query options part of an OData URL can be quite long, potentially exceeding the maximum length of URLs supported by components involved in transmitting or processing the request. One way to avoid this is wrapping the request in a batch request, which has the penalty of needing to construct a well-formed batch request body.</p>
-<p>An easier alternative for <code>GET</code> requests is to append <code>/$query</code> to the resource path of the URL, use the <code>POST</code> verb instead of <code>GET</code>, and pass the query options part of the URL in the request body.</p>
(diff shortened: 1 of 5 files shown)


HeikoTheissen (dismissed):

mikepizzo (changes requested):
We should add the use of the QUERY verb to the changes table in the introduction section, and also consider for the conformance section.

I can add the conformance proposal to #2230. I didn't find any conformance statements around the use of /$query.  Should we make both SHOULD in Intermediate for 4.02?  I think the HTTP QUERY method is too new to make it a must in any conformance level (it may not be supported in underlying HTTP libraries).

HeikoTheissen:
> I didn't find any conformance statements around the use of /$query. Should we make both SHOULD in Intermediate for 4.02?

Given that we have ignored `/$query` in all conformance statements so far, I suggest that we keep ignoring it in 4.02. Let's view it as a temporary workaround until the `QUERY` verb is widely adopted and rather demand support for `QUERY``
- for advanced 4.02 conformance
- or (intermediate) 4.03 conformance later.

HeikoTheissen (approved):

mikepizzo (approved):

Issue #2021, opened 2024-10-29
Insert (not update) a member in a PATCH collection


HeikoTheissen:
In PATCH collection requests, the control information `"@etag": "*"` can be used in a collection member to express that this shall be updated, but not inserted. This corresponds to the `If-Match: *` header.

What control information can be used if a collection member shall be inserted, not updated? This would correspond to the `If-None-Match: *` header. (It is not necessary to express arbitrary etag values here, only `*`.)

Issue #1965, opened 2024-07-10
Clarify wording for error in processing delta update


mikepizzo:
In [OData Version 4.02. Part 1: Protocol (oasis-open.org)](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#UpdateaCollectionofEntities) we say:

> If the [continue-on-error](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#Preferencecontinueonerrorodatacontinueonerror) preference has been specified and any errors occur in processing the changes, then a delta response MUST be returned regardless of the [return](https://docs.oasis-open.org/odata/odata/v4.02/csd01/part1-protocol/odata-v4.02-csd01-part1-protocol.html#Preferencereturnrepresentationandreturnminimal) preference and MUST contain at least the failed changes.

Since continue-on-error is a preference, shouldn't the service be allowed to fail the request entirely (without applying any changes) and return an error, within the continue-on-error preferenceapplied set to false?

I *think* what we meant was that, if the continue-on-error preference is *honored* and errors occur then the response must be returned.

ralfhandl:
Sounds reasonable

PR #2242, opened 2026-07-15
Added conformance requirements for 4.02

### LINKED ISSUE #2230: Define conformance requirements for 4.02

mikepizzo:
Fixes #2230 

Comments on addressed items:

# 4.02 CSDL Features

7. New path evaluation rules for annotations targeting annotations and external targeting via container (#575)
-we could say that 4.02 (intermediate?) must use the new external targeting. we could say that services should avoid container-bound targets with paths, because they were ambiguous in 4.0/4.01 (but clients should just silently ignore).  Note that, in this specification, we should not give a preference to one style of capability annotation over another, since this document should only be describing how to use/interpret the syntax of the CSDL.
- [x] Added that 4.02 clients MUST interpret according to clarified path evaluation rules.

8. Nested If without else part in collections (#326)
-_Propose_ Only affects annotations, so no conformance required.  If there is an equivalent 4.01 representation we could suggest that 4.02 services use the old syntax for downlevel clients (@HeikoTheissen will check to see if there is a 4.01 representation).
- [x] 2026/7/22: we don't believe there is an alternate representation that achieves the same purpose, in which case, there is no alternative guidance to give to interoperable clients.

# 4.02 Protocol
1. Responses that include errors MUST include the Preference-Applied header with continue-on-error set to true (#1965)
- [x] 2026/7/22: This is a clarification; no conformance statement is required.

# 4.02 JSON Format

2. Transient entities can be identifiable (#1928)
4.02 intermediate services SHOULD represent identifiable transient entities in response payloads using the pattern `odata:transient:{some-generated-identifier-unique-within-the-response}`, and if the transient entity cannot be re-read its `readLink` control information SHOULD have the `null` value.
- [x] 2026/7/22: Agreed to make SHOULD for all 4.02 JSON producers.

3. Control information "@etag": "" to prevent updates (#2021)
_Proposal_ 4.02 intermediate SHOULD support, advanced MUST support?
- [x] 2026/7/22: Already there for PUT, PATCH, DELETE, but not mention of special empty string for if none match.  So change for 4.02 conformance is to call out service SHOULD support empty string for if none match.

6.  Entities can be referenced by id or full set of key properties (#352)
_Proposal_: 4.02 intermediate services must support referencing by either id or full set of key properties. Interoperable clients must be prepared for full set of key properties in place of id in any response payload.
- [x] JSON doesn't have conformance levels. Made MUST for compliant 4.02 JSON consumers.

8. Allow common expressions in action payloads (#341)
_Proposal_: 4.02 advanced services SHOULD support expressions in non-binding parameters
- [x] JSON doesn't have conformance levels. Made SHOULD for 4.02 consumers

9. Atomicity group of prerequisite requests optional in dependsOn (#2150)
_Proposal_: 4.02 minimal services MUST NOT require atomicity group in dependsOn; 4.02 interoperable clients SHOULD specify
- [x] Made 4.02 services MUST NOT require, 4.02 producers SHOULD specify 

10. Ordering of properties in JSON Batch Request/Response objects (#351)
_Proposal_: 4.02 minimal ordered payloads and interoperable clients MUST follow ordering constraints in json batch request/response objects 
 - [x] Made MUST for 4.02 ordered payloads

# URL Conventions
1. POST ~/$query with Content-Type: application/x-www-form-urlencoded or application/json	https://github.com/oasis-tcs/odata-specs/issues/320 (#371)
 - [x] Made SHOULD in intermediate for both /$query and HTTP Query and both content-types

3. New overload for function matchespattern with flags (#441)
 - [x] New pattern - no conformance statement added

5. Nested query options can only appear once per expand item	#2004
 - [x] No Action performed. This appears to just be a clarification
 
9. Allow alternative $search syntax	(#293)
- [x] New pattern - no conformance added
 
# Additional Considerations
 - [x] Since there are no levels for JSON format, 4.02 JSON producers SHOULD always follow JSON Payload Ordering Constraints
 
- MUST support $orderby/$top/$skip in $select of collection-valued properties?
 2026/7/29: SHOULD be supported by advanced services in both $select and $expand
- [x] See below.  4.0 Advanced already defines $expand options that SHOULD/MUST be supported, and when we added $select options in 4.0 we added similar requirements for the $select options that a 4.01 Advanced service SHOULD/MUST supported.
 
Should we remove $search from should/must support for intermediate/advanced services?
 - [x] No. Can't really make the change since a 4.02 intermediate/advanced must be a 4.01 intermediate/advanced.


odata-json-format/23 Conformance.md
```diff
@@ -34,47 +32,67 @@ In order to be a conforming consumer of the OData JSON format, a client or servi
    5. MUST accept the string values `-INF`, `INF`, and `NaN` for decimal values with floating scale
    6. MUST be prepared to handle related entities inline within a delta payload as well as a nested delta representation for the collection
    7. MUST be prepared to handle decimal values written in exponential notation
+9. MAY be a conforming consumer of the OData 4.02 JSON format, for payloads with an `OData-Version` header value of `4.02`.
+   1. MUST support referencing entities by either id or full set of key properties
+   2. MUST NOT require the atomicity group also be included in `dependsOn` for prerequisit requests that are in an atomicity group
```
ralfhandl:
```suggestion
   2. MUST NOT require the atomicity group also be included in `dependsOn` for prerequisite requests that are in an atomicity group
```

odata-json-format/23 Conformance.md
```diff
@@ -34,47 +32,67 @@ In order to be a conforming consumer of the OData JSON format, a client or servi
...
     6. MAY return decimal values written in exponential notation
     7. MAY represent numeric value exceptions for decimal values with floating scale using the string values `-INF`, `INF`, and `NaN`
+12. MAY support generating OData 4.02 JSON compliant payloads for requests with an `OData-Version` header value of `4.02`.
+    1. MUST NOT percent-encode the fragment portion of a Context URL
+    2. MUST order `@type` control information, if specified, immediately following `@removed` in a deleted entity for ordered payloads
+    3. MUST return the `requestId` as the first property in a JSON Batch Request or Response Object and `body`, if specified, as the last property for ordered payloads
+    4. SHOULD always follow [JSON Payload Ordering Constraints](PayloadOrderingConstraints), regardless the presence of the `odata.streaming` format parameter
+    5. SHOULD represent identifiable transient entities using the pattern `odata:transient:{unique-identifier}`, and with an explicit `null` value for the `readlink` if the transient entity cannot be re-read
+    6. SHOULD include the atomicity group in `dependsOn` for prerequisit requests that are in an atomicity group
```
ralfhandl:
```suggestion
    6. SHOULD include the atomicity group in `dependsOn` for prerequisite requests that are in an atomicity group
```

odata-json-format/23 Conformance.md
mikepizzo: fixed - thanks.

mikepizzo: Fixed -- thanks

odata-json-format/23 Conformance.md
mikepizzo: fixed- thanks.

odata-protocol/12 Conformance.md
```diff
@@ -383,6 +391,23 @@ service:
 1. MUST conform to the [OData 4.02 Minimal Conformance Level](#OData402MinimalConformanceLevel)
 2. MUST conform to the [OData 4.01 Intermediate Conformance
 Level](#OData401IntermediateConformanceLevel)
+3. MUST support `select` with an empty select list
```
HeikoTheissen:
```suggestion
3. MUST support `$select` with an empty select list
```

HeikoTheissen (dismissed):

mikepizzo:
regarding:

 
>> -  MUST support $orderby/$top/$skip in $select of collection-valued properties?
>> - [ ] Not Done. 2026/7/29 SHOULD be supported by advanced services in both $select and $expand

It turns out, we do require certain $expand options in OData 4.0 Advanced:
~~~json
9. MUST support `$expand` ([section ##SystemQueryOptionexpand])
   1. MUST support returning references for expanded properties
   2. MUST support `$filter` on expanded collection-valued properties
   3. MUST support cast segment in expand with derived types
   4. SHOULD support `$orderby` with `asc` and `desc` on expanded
collection-valued properties
   5. SHOULD support `$count` on expanded collection-valued properties
   6. SHOULD support `$top` and `$skip` on expanded collection-valued
properties
   7. SHOULD support `$search` on expanded collection-valued properties
   8. SHOULD support `$levels` for recursive expand ([section ##ExpandOptionlevels])
   9. MAY support `$compute` on expanded properties
   ~~~

and 4.01 Intermediate SHOULD support nested query options in $select:
~~~json
9. SHOULD support nested query options in `$select`
~~~

and a 4.01 Advanced service MUST support the same set of $select options as 4.0 Advanced $expand options:
~~~json
5. MUST support nested options in `$select`
   1. MUST support `$filter` on selected collection-valued properties
   2. SHOULD support `$orderby` with `asc` and `desc` on selected
collection-valued properties
   3. SHOULD support the `$count` on selected collection-valued
properties
   4. SHOULD support `$top` and `$skip` on selected collection-valued
properties
   5. SHOULD support `$search` on selected collection-valued properties
 ~~~

So, unless we want to make some of these SHOULDs a MUST (which was not in line with our discussion on 2026/7/20) then I think we have it covered.

HeikoTheissen (approved):

PR #2246, opened 2026-07-19
Add conformance statements for empty select/expand lists

### LINKED ISSUE #2243: Add a way to suppress $select and $expand
#2031 added the clarification that deep inserts must be expanded to the level of the request.  This brought up the fact that there is currently no way to suppress expanding navigation properties that would otherwise automatically be expanded (either through AutoExpand or through protocol requirements).

#2031 contemplated various different ways to suppress auto-expansion, and https://github.com/oasis-tcs/odata-vocabularies/pull/295 proposed the addition of a new token annotation to supppress.

In reviewing the change proposal, the TC agreed that a more natural way to represent suppressing default expand (and select) would be through an empty select/expand list.

mikepizzo:
Fixes #2243

Add conformance statements for empty select/expand lists.


Hubert-Heijkers (approved):
Sounds good to me! Maybe having the MUST here is perhaps good enough to retain the MAY in protocol and url-conventions.

PR #2261, opened 2026-08-05
Add $key to select all keys

### LINKED ISSUE #2257: What is the ContextUrl for an empty select list?
In #2243 we added the ability to specify an empty select or expand list (meaning, don't select or expand by default).

If there are no navigation properties suffixed with parens  in a the {select-list} of the contextUrl, then it means no navigation properties were expanded, so that's fine.

But, a navigation property suffixed by an empty expand list means no select or expand (so default selected properties):
`If the expanded navigation property does not contain a nested $select or $expand, then the expanded property is suffixed with empty parentheses.`

What if no properties of the root or expanded entity are selected?
 
If the service returns key properties, instead of ids, then the service would return those key properties in the select-list, so it would be fine, but what if the service just returns odata ids?

Note that the current spec says that, if the {select-list} of the contextUrl includes only navigation properties suffixed with parens (empty or otherwise), then the default set of properties are returned:
`If the context URL includes only expanded navigation properties (i.e., only navigation properties suffixed with parentheses), then all structural properties are implicitly selected (same as if there were no properties listed in the select-list).`

So, then, how do you represent that no properties are selected and only ids (not key properties) are returned?

mikepizzo:
remove empty $select list

Fixes #2257

### DIFF
--- a/odata-protocol/11 Data Service Requests.md
+++ b/odata-protocol/11 Data Service Requests.md
@@ -334,10 +334,8 @@ alias of the schema in order to specify all operations defined in the
 schema. Only aliases defined in the metadata document of the service can
 be used in URLs.
 
-OData 4.02 and greater services MAY support an empty select list to indicate
-that no structural properties are requested. The service still returns expanded
-navigation and stream properties, along with required metadata and/or key properties
-as required by the particular format.
+OData 4.02 and greater services MAY support `$key` in the select list to indicate
+that all key properties be included in the response.
 
 ::: example
 Example ##ex: request only the `Rating` and `ReleaseDate` for the matching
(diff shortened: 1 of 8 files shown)


odata-protocol/11 Data Service Requests.md
```diff
@@ -334,10 +334,8 @@ alias of the schema in order to specify all operations defined in the
 schema. Only aliases defined in the metadata document of the service can
 be used in URLs.
 
-OData 4.02 and greater services MAY support an empty select list to indicate
-that no structural properties are requested. The service still returns expanded
-navigation and stream properties, along with required metadata and/or key properties
-as required by the particular format.
+OData 4.02 and greater services MAY support `$key` in the select list to indicate
+that all key properties be included in the response.
```
HeikoTheissen, line 338: Should we say that this applies only if the resource path addresses entities (not complex instances)?

Hubert-Heijkers: I'd argue that is kind of implicit as only entity types can have keys to begin with. Are we that explicit whenever we talk about anything related to keys?

HeikoTheissen (approved):

PR #2267, opened 2026-08-13
Prepare Version 4.02 Committee Specification Draft 02

HeikoTheissen:


odata-protocol/meta.yaml
```diff
@@ -18,12 +18,15 @@ additionalArtifacts:
...
+    path: syntax-diagram/odata-abnf-diagrams.html
     branch: v4.02-release-candidate
-OData-ABNF: https://oasis-tcs.github.io/odata-abnf/odata-abnf-diagrams.html
-OData-VocCore: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md
-OData-VocCap: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md
-OData-VocAggr: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Aggregation.V1.md
+  - repo: odata-abnf
+    path: syntax-diagram/railroad-diagram.css
+    branch: v4.02-release-candidate
+OData-ABNF: https://docs.oasis-open.org/odata/odata/v4.02/csd02/syntax-diagram/odata-abnf-diagrams.html
```
HeikoTheissen: I would prefer railroad diagrams from https://oasis-tcs.github.io, but we cannot serve the v4.02-release-candidate branch from GitHub pages.

HeikoTheissen:
In the ZIP files that contain ABNF, there are two subfolders `abnf` and `syntax-diagram` on the same level, reflecting the folder structure in https://github.com/oasis-tcs/odata-abnf. Should we make `syntax-diagram` a subfolder of `abnf`?

Made `syntax-diagram` a subfolder of `abnf`.

odata-protocol/Appendix.md
```diff
@@ -213,7 +213,9 @@ The following individuals were members of the OASIS OData Technical Committee du
 | :--- | :--- | :--- | :--- |
 |Committee Specification Draft 01|2024-02-28|Michael Pizzo<br>Ralf Handl<br>Heiko Theißen| 
 Import material from OData Version 4.01 Part 1: Protocol 
-<br>Changes listed in [section ##ChangesfromEarlierVersions]|
+<br>Changes listed in section 1.1|
+|$$$subtitle$$$|$$$pubdateISO$$$|Michael Pizzo<br>Ralf Handl<br>Heiko Theißen| 
+Changes listed in [section ##ChangesfromEarlierVersions]|
```
HeikoTheissen, line 222: Is this entry in the Version History OK? Similar in all other spec documents.

odata-url-conventions/meta.yaml
```diff
@@ -5,15 +5,15 @@ subtitle: Committee Specification Draft 02
...
+pubdateISO: '2026-08-19'
+copyright: © OASIS Open 2026
 archivename: odata-v4.02-csd02
 partname: part2-url-conventions
-OData-ABNF: https://oasis-tcs.github.io/odata-abnf/odata-abnf-diagrams.html
-OData-VocCore: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Core.V1.md
-OData-VocCap: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Capabilities.V1.md
-OData-VocAggr: https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies/Org.OData.Aggregation.V1.md
+OData-ABNF: https://docs.oasis-open.org/odata/odata/v4.02/csd02/syntax-diagram/odata-abnf-diagrams.html
+OData-VocCore: https://github.com/oasis-tcs/odata-vocabularies/blob/v4.02-release-candidate/vocabularies/Org.OData.Core.V1.md
```
ralfhandl, line 15: Using the branch name means we have to keep it forever.

If we don't want to use `main`, we should use `v4.02-csd02` and protect that branch against changed and deletion.

odata-protocol/meta.yaml
ralfhandl: I prefer an immutable HTML file next to the immutable spec files.

odata-protocol/Appendix.md
ralfhandl: Problem here is that section 1.1 lists differences between _versions_ (V4.0, V4.01, V4.02, ...) and the table in appendix D lists differences between _revisions_ (CSD01, CSS02, ...) of the same version, except for the very first line, which lists differences between the last revision of the previous version and the first revision of the current version. Only for the first revision of a new version this is identical to section 1.1.

This means that in revision CSD02 the first line of the revision table cannot point to section 1.1 of the same document, neither can the second line.

We could split section 1.1 into subsections for each revision. 

Or we could add a revision column to the table in section 1.1.

odata-url-conventions/meta.yaml
HeikoTheissen: I now use the vocabularies document published at OASIS instead: https://docs.oasis-open.org/odata/odata-vocabularies/v4.0/csd02/odata-vocabularies-v4.0-csd02.html with csd02 replaced by csd03.

odata-protocol/meta.yaml
HeikoTheissen: Where exactly? For now, it would be at https://docs.oasis-open.org/odata/odata/v4.02/csd02/abnf/syntax-diagram/odata-abnf-diagrams.html.

odata-protocol/Appendix.md
HeikoTheissen: Added the revision column.

odata-protocol/meta.yaml
ralfhandl: Sounds good. 

We should check with TC Admin beforehand which folder names/structure they prefer.

HeikoTheissen: The ABNF "additional artifact" is given as
> ABNF components: OData ABNF Construction Rules Version 4.02 and OData ABNF Test Cases Version 4.02. https://docs.oasis-open.org/odata/odata/v4.02/csd02/abnf/.

Then `syntax-diagram` should be subfolder of that.

HeikoTheissen:
Replaced by #2269

PR #2270, opened 2026-08-20
Compact JSON Format

HeikoTheissen:


odata-compact-json-format/5 Positional Representation.md
```diff
@@ -0,0 +1,269 @@
...
+of the positional property list; the two requests differ only in the
+order of the `$select` items and produce different positional
+representations of the same entity
+```
+GET ~/Customers('ALFKI')?$select=Name,ID
+{
+  "@context": "$metadata#Customers(Name,ID)/$entity",
+  "_": ["Alfreds Futterkiste", "ALFKI"]
+}
+```
```
HeikoTheissen: Suggest to write such examples as two code blocks:
````suggestion
```
GET ~/Customers('ALFKI')?$select=Name,ID
```
```json
{
  "@context": "$metadata#Customers(Name,ID)/$entity",
  "_": ["Alfreds Futterkiste", "ALFKI"]
}
```
````

Comment (Hubert-Heijkers) [reply odata-compact-json-format/5 Positional Representation.md]: I'll tell Claude to get it's act together on this one;-!

I'm reviewing the whole doc myself again as well, got a number of little changes/open points as a follow up (some of which I might address myself before Wednesday).

File: odata-compact-json-format/6 Annotations and Control Information.md [hunk 2]
```diff
@@ -0,0 +1,271 @@
...
+
+A wrapper object that carries no value denotes a property or instance
+that has no value, as distinct from one whose value is null. The empty
+wrapper object `{}` therefore denotes "no value and no annotations"; see
+[section ##DerivedTypes].
+
+The name of the value in a wrapper object is `_`.
+
+At the root of the message body the name `value` is also recognized, with
+the same meaning, for compatibility with [OData-JSON](#ODataJSON) as
```
HeikoTheissen: But [OData-JSON] allows the `value` name only for a collection of entities,
```json
{"@context": "...",
 "value": {...}}
```
is not a valid representation of a single entity.

odata-compact-json-format/1 Introduction.md
```diff
@@ -0,0 +1,184 @@
...
+the format defined by [OData-JSON](#ODataJSON) and once in the format
+defined by this document, to make the correspondence between them
+apparent. Where an example shows only one of the two, the other is not
+relevant to the point being made.
+
+All other text is normative unless otherwise labeled.
+
+-------
+
+# ##sec Compact JSON Format Design
```
HeikoTheissen: Since this starts section 2, all subsequent file names are off by 1. ("2 Requesting the ..." should be "3 Requesting the ..." and so on.)

odata-compact-json-format/1 Introduction.md
```diff
@@ -0,0 +1,184 @@
...
+repetition of property names, once per property per instance, dominates
+that volume in exactly the payloads where it hurts most: large collections
+of entities and large collections of complex values.
+
+The approach taken in this format, inspired by traditional rowset
+interfaces, is to represent a structured instance as a JSON *array* rather
+than a JSON object, so that a property name is transmitted at most once
+per payload instead of once per instance. JSON arrays preserve the order
+of their items, whereas the order of the name/value pairs of a JSON object
+is not significant; this format uses that property of arrays to convey,
```
ralfhandl:
```suggestion
is not significant; this format uses that feature of arrays to convey,
```

odata-compact-json-format/6 Annotations and Control Information.md
Hubert-Heijkers: Correct @HeikoTheissen, there is an inherent conflict here, updated the sections around this for further clarification.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+header.
+
+The names and values of the format parameters are case-insensitive.
+
+A service that does not support the compact JSON format MUST NOT return a
+compact payload. Because the `compact` format parameter selects a payload
+shape that a receiver unaware of this specification will misinterpret, a
+service MUST NOT return a compact response unless the client requested it
+with `compact=true`, and MUST return `406 Not Acceptable` if the client
+requests `compact=true` and the service does not support it.
```
mikepizzo, line 25: If regular JSON is a subset of compact JSON, why must the service return not acceptable if the client requests compact=true and the service doesn't support it.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+
+A client MUST NOT send a compact request body unless it has established
+that the service supports the compact JSON format, either from the
+[`Capabilities.SupportedFormats`](#AdvertisingSupport) annotation or from
```
mikepizzo, line 29: how is support for compact json specified in supported formats?  is it just application/json;compact=true?

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+out-of-band knowledge. A service that receives a request body labeled with
```
mikepizzo, line 30: An existing service won't necessarily know to return this.  We should just say that "A service that does not support compact=true returns `415 Unsupported Media Type` " and drop the MUST.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
mikepizzo: Note: service should not claim to support compact json if all responses are standard json, but a service should always be allowed to return regular json if the client requests compact=true.

odata-compact-json-format/6 Structured Instances.md
```diff
@@ -0,0 +1,238 @@
...
+::: example
+Example ##ex: `PreferredContact` is declared by `Model.VipCustomer` only;
+the first entity is not a `VipCustomer` and its position for
+`PreferredContact` conveys "not applicable"
+```
+GET ~/Customers?$select=ID,Model.VipCustomer/PreferredContact
+{
+  "@context": "$metadata#Customers(ID,Model.VipCustomer/PreferredContact)",
+  "_": [
+    ["ALFKI", {}],
```
mikepizzo: alternatively, we could identify the set of relevant properties per row based on the type.  That is, since this is the base type I know that only "ALFKI" applies, so I would only look for one property.

odata-compact-json-format/6 Structured Instances.md
```diff
@@ -0,0 +1,238 @@
...
+::: example
+Example ##ex: a payload in which one instance carries an unselected
+dynamic property and therefore falls back to the representation defined
+by [OData-JSON](#ODataJSON)
+```json
+{
+  "@context": "$metadata#Customers(ID,Name)",
+  "_": [
+    ["ALFKI", "Alfreds Futterkiste"],
+    { "ID": "ANATR", "Name": "Ana Trujillo", "Nickname": "Ana" }
```
mikepizzo: alternatively, could I do:
    { "_": ["ANATR", "Ana Trujillo"], "Nickname": "Ana" }

odata-compact-json-format/7 Annotations and Control Information.md
```diff
@@ -0,0 +1,271 @@
...
+  the object *containing* the property;
+- control information follows the same two patterns, with a reserved term
+  name, for example `@id` and `Orders@count`.
+
+Neither pattern is available in a positional representation: an array has
+no name/value pairs, and the containing instance may itself be an array.
+This format therefore introduces a single construct, the *wrapper
+object*, which is used uniformly wherever a value would otherwise appear.
+
+## ##subsec The Wrapper Object
```
mikepizzo: Organizationally, it's a bit confusing to introduce the wrapper object *after* it's been used a number of times to describe things like annotations, control information, and dynamic properties.

odata-compact-json-format/1 Introduction.md
Hubert-Heijkers: Created a file per section and fixed up the numbering accordingly

HeikoTheissen: But you did not push that change yet, right?

odata-compact-json-format/6 Structured Instances.md
Hubert-Heijkers: Had a good heart-to-heart with Claude/Fable and after looking at the definition of the wrapper a bit more, discussion what property name to use for the positional representation after agreeing that `_` wasn't a good choice (discussed "", "@" and "$" mainly), adding this alternative to the mix settled it, `$` it is (what we propose) ;-)

odata-compact-json-format/6 Structured Instances.md
Hubert-Heijkers: Worked this into the pending updated proposal as well!

odata-compact-json-format/3 Requesting the Compact JSON Format.md
Hubert-Heijkers: It would only be if the client's Accept list ONLY allows for `application/json; compact=true` and NOT `application/json`. This to me is standard content-type negotiation saying only if `application/json; compact=true` is in the Accept list you are allowed to choose it and if it's not in there and it's the only acceptable content type then you MUST return `406 Not Acceptable` (which is the correct error according to the content negotiation specification).

odata-compact-json-format/3 Requesting the Compact JSON Format.md
Hubert-Heijkers: That's what I'd expect indeed. This is what we have today already:

```
"@Capabilities.SupportedFormats": [
                "application/json",
                "application/json;odata.metadata=full",
                "application/json;odata.metadata=minimal",
                "application/json;odata.metadata=none",
                "application/json;IEEE754Compatible=false",
                "application/json;IEEE754Compatible=true",
                "application/json;odata.streaming=true",
                "application/json;IEEE754Compatible=false;odata.streaming=true",
                "application/json;IEEE754Compatible=true;odata.streaming=true",
                "application/json;odata.metadata=full;IEEE754Compatible=false",
                "application/json;odata.metadata=minimal;IEEE754Compatible=false",
                "application/json;odata.metadata=none;IEEE754Compatible=false",
                "application/json;odata.metadata=full;IEEE754Compatible=true",
                "application/json;odata.metadata=minimal;IEEE754Compatible=true",
                "application/json;odata.metadata=none;IEEE754Compatible=true",
                "application/json;odata.metadata=full;odata.streaming=true",
                "application/json;odata.metadata=minimal;odata.streaming=true",
                "application/json;odata.metadata=none;odata.streaming=true",
                "application/json;odata.metadata=full;odata.streaming=true",
                "application/json;odata.metadata=minimal;odata.streaming=true",
                "application/json;odata.metadata=none;odata.streaming=true",
                "application/json;odata.metadata=full;IEEE754Compatible=false;odata.streaming=true",
                "application/json;odata.metadata=minimal;IEEE754Compatible=false;odata.streaming=true",
                "application/json;odata.metadata=none;IEEE754Compatible=false;odata.streaming=true",
                "application/json;odata.metadata=full;IEEE754Compatible=true;odata.streaming=true",
                "application/json;odata.metadata=minimal;IEEE754Compatible=true;odata.streaming=true",
                "application/json;odata.metadata=none;IEEE754Compatible=true;odata.streaming=true"
            ]
```

Isn't graph representing the combination of parameters supported that way as well @mikepizzo?

odata-compact-json-format/3 Requesting the Compact JSON Format.md
Hubert-Heijkers: Think you are mixing something up here @mikepizzo, `415 Unsupported Media Type` would be returned by the server if the server DOES NOT support compact JSON and a client sent it a request with a request body containing compact JSON and as such a content type header containing `application/json; compact=true`.

odata-compact-json-format/1 Introduction.md
Hubert-Heijkers: Yeah, I see, property appears twice with two different meanings but 'feature' is a bit vague (and faintly marketing-flavored for a normative document according to Claude <LOL>) so changed it to read:

```
The items of a JSON array are ordered, whereas the name/value pairs of a JSON object are not; this format relies on that ordering to convey, by position, which value belongs to which property.
```

odata-compact-json-format/7 Annotations and Control Information.md
Hubert-Heijkers: Agreed @mikepizzo, moved it into section 4 Compact Representation as it is the combination of the positional representation and these wrapper objects that allows us to structure these compact payloads.

Hubert-Heijkers:
@HeikoTheissen, @ralfhandl, @mikepizzo, I think I applied all the topics I wanted to address as well as dealt with all the comments and things that came up during our chat on Wednesday.
The open issues section is update to reflect what topics are left to cover but, whilst I have ideas on those and some are even expressed in there as well, these need further discussion before I want to apply them.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
ralfhandl, line 25:
```suggestion
only accepts `compact=true` and the service does not support it.
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+Everywhere else, `value` is not the name of a wrapper object's value. In
+particular, where [OData-JSON](#ODataJSON) represents the message body as
+the instance itself -- for a single entity, a single complex value, or a
+single entity reference -- a name/value pair named `value` in that message
+body is a *property* named `value`, and a receiver MUST NOT read it as the
+value of a wrapper object. A single entity or complex value represented
+positionally at the root of the message body therefore uses `$`.
+
+Producers of compact payloads SHOULD use `$` wherever this document
```
ralfhandl, line 348: What do senders gain by being able to choose `value` instead of `$`?

In which cases would that choice be helpful?

odata-compact-json-format/8 Request Payloads.md
```diff
@@ -0,0 +1,222 @@
...
+The select-list of the `context` control information carries this
+distinction: the positional property list *is* the list of properties to
+be updated. A property that is not in the positional property list is
+absent from the request in the sense of
+[OData-Protocol](#ODataProtocol) and is left unchanged. A property that
+is in the positional property list with the value `null` is set to null.
+
+A `PATCH` request body that does not contain the `context` control
+information therefore specifies a value for every structural property of
+the type, which is rarely the intent. Clients SHOULD include the `context`
```
ralfhandl, line 100: Lines 22-23 say

>  compact request body MUST contain the
> [`context`](#ControlInformationcontext) control information

Why relax that here?

odata-compact-json-format/8 Request Payloads.md
```diff
@@ -0,0 +1,222 @@
...
+## ##subsec Bind Operations
+
+A bind operation relates a new or updated entity to an existing entity.
+[OData-JSON](#ODataJSON) represents it as a name/value pair
+`Property@bind` or `Property@id` on the containing object.
+
+In a compact request body the navigation property occupies a position in
+the positional property list, and the bind operation is carried in a
+[wrapper object](#wrapperobject) at that position, as a property
+annotation, following [section ##PropertyAnnotations].
```
ralfhandl, line 166:
```suggestion
[wrapper object](#wrapperobject) at that position, as control
information, following [section ##PropertyAnnotations].
```

ralfhandl (commented):
A few questions and minor things

odata-compact-json-format/1 Introduction.md
```diff
@@ -0,0 +1,108 @@
+
+-------
+
+# ##sec Introduction
+
+The OData protocol is comprised of a set of specifications for
+representing and interacting with structured content. The core
+specification for the protocol is in [OData-Protocol](#ODataProtocol);
+this document is an extension of the core protocol. Representations for
```
mikepizzo, line 9: extension of the core protocol, or of the JSON format?

odata-compact-json-format/1 Introduction.md
```diff
@@ -0,0 +1,108 @@
...
+OData requests and responses using the JavaScript Object Notation (JSON),
+see [RFC8259](#rfc8259), are defined in [OData-JSON](#ODataJSON).
+
+This document defines a *compact* JSON format: a lossless, alternative
+representation of the same information, whose prime aim is to minimize
+the uncompressed size of OData request and response payloads.
```
mikepizzo, line 15:
```suggestion
the uncompressed size of OData JSON request and response payloads.
```

odata-compact-json-format/1 Introduction.md
```diff
@@ -0,0 +1,108 @@
...
+
+Processing large volumes of JSON is expensive, largely due to its
+verbosity. Compression on the wire takes most of that inefficiency away
+during transport, but the sender producing and the receiver consuming the
+uncompressed payload still deal with the full volume of JSON text. The
+repetition of property names, once per property per instance, dominates
+that volume in exactly the payloads where it hurts most: large collections
+of entities and large collections of complex values.
+
+The approach taken in this format, inspired by traditional rowset
```
mikepizzo, line 25: I'm not sure what is implied by "traditional rowset interfaces".  I think of this more similar to CSV, but not sure we need the comment at all.

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+
+# ##sec Compact JSON Format Design
+
+[OData-JSON](#ODataJSON) extends JSON by defining general conventions for
+name/value pairs that annotate a JSON object, property or array, and a set
+of canonical name/value pairs for control information such as ids, types
+and links.
+
+This format extends [OData-JSON](#ODataJSON) further by allowing a
+structured instance -- an entity or a complex value -- to be represented as
```
mikepizzo, line 12: specifically, the properties of a structured instance, no?

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+## ##subsec Design Principles
+
+The following principles govern this format. They are stated here because
+they explain, and constrain, the rules in the remainder of this document.
+
+1. [*Compact JSON is a superset of the OData JSON format.*]{id=supersetprinciple}
+   Every payload that is valid according
+   to [OData-JSON](#ODataJSON) is also a valid compact JSON payload. A
+   receiver that accepts compact JSON therefore accepts strictly more
+   payloads than a receiver that accepts only [OData-JSON](#ODataJSON),
```
mikepizzo, line 25: the point is not that it accepts more, but that it accepts any valid OData JSON Payload.
```suggestion
   to [OData-JSON](#ODataJSON) is also a valid compact JSON payload. A
   receiver that accepts compact JSON therefore also accepts [OData-JSON](#ODataJSON),
```
although this may be redundant with the statement that "Compact JSON is a superset of the OData JSON Format."

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+   and a sender is never forced to use a positional representation where
+   it is inconvenient or impossible.
+
+2. *The saving is in the repetition.* A positional representation removes
```
mikepizzo, line 29: nit: isn't the savings in *avoiding* the repetition?

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+   property names from a payload; the mapping from positions to properties
```
mikepizzo, line 30: the property names are still in the context url...
```suggestion
   redundant property names from a payload; the mapping from positions to properties
```

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+   is transmitted once. Consequently the benefit grows with the number of
+   instances that share a positional property list, and is negligible or
+   negative for a payload containing a single instance. This format
+   therefore never *requires* a positional representation.
```
mikepizzo, line 34: I don't know that the optionality of the positional representation follows directly from the fact that benefit is negligible or negative for a single instance. I think there are a number of benefits to the design principle that you compact JSON is a superset of JSON, and this is just one of them.


```suggestion
   negative for a payload containing a single instance. As the Compact JSON format
   never *requires* a positional representation, producers can fall back to the standard
   representation where convenient or more concise.
```

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+
+4. *Requests and responses are treated alike.* Large collections occur in
+   request bodies -- of `POST`, `PUT` and `PATCH` requests -- as well as in
+   response bodies, and the saving is the same in both directions. Where
+   requests and responses must differ, the difference is called out
+   explicitly.
+
+::: example
+Example ##ex_superset: because of the [superset
+principle](#supersetprinciple), a sender that cannot produce a positional
```
mikepizzo, line 50: throughout -- I think we generally use "producer" rather than "sender".

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+representation for a particular instance may fall back to the
+representation defined by [OData-JSON](#ODataJSON) for that instance
+alone, within an otherwise positional payload:
```
mikepizzo, line 53: I would strike "alone" -- they can do it for an arbitrary number within the payload.
```suggestion
 within an otherwise positional payload:
```

odata-compact-json-format/2 Compact JSON Format Design.md
mikepizzo, line 50: cannot, or choses not to

odata-compact-json-format/2 Compact JSON Format Design.md
```diff
@@ -0,0 +1,77 @@
...
+## ##subsec Relationship to the OData JSON Format
+
+This document specifies only where the compact JSON format deviates from
+[OData-JSON](#ODataJSON). It is not a self-contained format
+specification: for the representation of primitive values, for the
+definition and meaning of control information, for URL and relative URL
+handling, and for everything else it does not mention, the rules of
+[OData-JSON](#ODataJSON) apply unchanged.
+
+In particular, this format changes only *how a structured instance is
```
mikepizzo, line 75: is it valid to say that it changes only "how the positional properties of a structured instance are laid out"?

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+# ##sec Requesting the Compact JSON Format
+
+The compact JSON format can be requested using the `$format` query option
+in the request URL with the media type `application/json` followed by the
+`compact=true` format parameter, optionally followed by other format
+parameters.
+
+Alternatively, this format can be requested using the `Accept` header with
+the media type `application/json` followed by the `compact=true` format
+parameter, optionally followed by other format parameters.
```
mikepizzo, line 13: 1. I would reverse the order of these. Content-negotiation is the standard HTTP way to request a particular format, so we should describe that first, and then the OData-specific query option mechanism.
2. "followed by the compact=true format parameter, optionally followed by other format parameters" implies an ordering -- that compact=true must be the first format parameter. I don't believe the format parameters are (or should be) ordered.

```suggestion
The compact JSON format can be requested using the `Accept` header with
the media type `application/json` with the `compact=true` format
parameter.

Alternatively, this format can be requested using the `$format` query option
in the request URL with the media type `application/json` with the
`compact=true` format parameter.

```

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+
+If specified, `$format` overrides any value specified in the `Accept`
+header.
+
+The names and values of the format parameters are case-insensitive.
```
mikepizzo, line 18: Maybe move this to before "If specified..." since it applies to both?  Also, this is already defined in the MIME spec.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
mikepizzo, line 25: I'm not sure what "A service that does not support a compact JSON format MUST NOT return a compact payload" means -- isn't this self-evident?

Also, since compact JSON is a superset of JSON, I think it should be legal for a client to request compact json and the service to return standard JSON. If we want to know if the service supports compact json, we should instead use the appropriate capability to discover which formats are supported by the service.

Also, content-negotiation is a bit more complicated than saying that a service must return not acceptable if the client requests a format that it doesn't support -- since the client can specify multiple acceptable formats, the service only returns not acceptable if it doesn't support *any* of the requested formats.

```suggestion
Because the `compact=true` format parameter selects a payload
shape that a receiver unaware of this specification will misinterpret, a
service MUST NOT return a compact response unless the client requested it
with `compact=true`.
```

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+service MUST NOT return a compact response unless the client requested it
+with `compact=true`, and MUST return `406 Not Acceptable` if the client
+requests `compact=true` and the service does not support it.
+
+A client MUST NOT send a compact request body unless it has established
+that the service supports the compact JSON format, either from the
+[`Capabilities.SupportedFormats`](#AdvertisingSupport) annotation or from
+out-of-band knowledge. A service that receives a request body labeled with
+`compact=true` that it does not support MUST reject the request with
+`415 Unsupported Media Type`.
```
mikepizzo, line 32: I'm not sure we can say that a service that requests a request body with `compact=true` must reject with unsupported media type; doing so adds this as a requirement to existing services and [rfc2045](https://datatracker.ietf.org/doc/html/rfc2045#page-5) currently says: " MIME implementations must ignore any parameters whose names they do not recognize."

We could say that the behavior of a service that receives a request body labeled with `compact=true` that it does not support is undefined, and/or that services SHOULD (MAY?) reject the request with `415 Unsupported Media Type`.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+## ##subsec Format Parameters
+
+The format parameters defined in [OData-JSON](#ODataJSON) apply to the
+compact JSON format with the meaning defined there, subject to the
+following.
+
+The `streaming` parameter is meaningful and MAY be specified. A positional
+representation is inherently ordered, and a compact payload that meets the
+[payload ordering constraints](#PayloadOrderingConstraints) MUST include
+`streaming=true`.
```
mikepizzo, line 43: MUST?  if the payload happens to follow ordering constraints but does not specify streaming=true, is it non-compliant? I think this needs to be SHOULD...

```suggestion
[payload ordering constraints](#PayloadOrderingConstraints) SHOULD include
`streaming=true`.
```

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+`415 Unsupported Media Type`.
+
+## ##subsec Format Parameters
+
+The format parameters defined in [OData-JSON](#ODataJSON) apply to the
+compact JSON format with the meaning defined there, subject to the
+following.
+
+The `streaming` parameter is meaningful and MAY be specified. A positional
+representation is inherently ordered, and a compact payload that meets the
```
mikepizzo, line 41: I think the statement that "A positional representation is inherently ordered" is misleading.  Yes, the properties are ordered, but the streaming=true is parameter is more about control information ordering.

Note that OData-defined structures, like a batch request/response objects, are an interesting case -- their order is defined by odata for ordered payloads, but there are no properties in a contextUrl. can such properties be optimized in compact json?  I guess I'll read further to find out...

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+`minimal` or `none`. Note that `metadata=none` does not remove the
+[`context`](#ControlInformationcontext) control information from a compact
+payload: the context URL determines the [positional property
+list](#positionalpropertylist) and is therefore required for the payload
+to be interpretable at all. See [section ##ControlInformationcontext].
+
+The value `metadata=full` is NOT RECOMMENDED with `compact=true`. Full
+metadata requires control information to be present for every instance,
+which forces every instance into a [wrapper object](#wrapperobject) and
+defeats the purpose of the format. Services MAY reject the combination
```
mikepizzo, line 58: last week we I think we discussed being able to specify control information (and dynamic properties) as name/value pairs withing the wrapper object along with known properties in the array, which would still provide benefit to a compact representation of full metadata.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
```diff
@@ -0,0 +1,78 @@
...
+combinations of supported format parameters, including the `compact=true`
+format parameter if the compact JSON format is supported by the service.
+
+::: example
+Example ##ex: a service advertising support for both the format defined in
+[OData-JSON](#ODataJSON) and the compact JSON format
+```json
+"@Capabilities.SupportedFormats": [
+  "application/json;IEEE754Compatible=true",
+  "application/json;IEEE754Compatible=true;compact=true"
```
mikepizzo, line 75: I think we want to discourage IEE754Compatible unless absolutely required. I would use streaming=true instead.

```suggestion
  "application/json;",
  "application/json;streaming=true",
  "application/json;compact=true",
  "application/json;streaming=true;compact=true"
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
+
+-------
+
+# ##sec Compact Representations
+
+This section defines the two representations that this format adds to
+[OData-JSON](#ODataJSON): the *positional representation*, which conveys a
```
mikepizzo, line 7:
```suggestion
[OData-JSON](#ODataJSON): the *positional representation*, which conveys the properties of a
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+structured instance as a JSON array, and the *wrapper object*, which
+conveys whatever needs a name in a place where a positional representation
+has no room for one. Everything else in this document is expressed in
+terms of these two.
+
+## ##subsec Positional Representation
+
+A structured instance -- an entity or a complex value -- MAY be represented
+as a JSON array instead of a JSON object. Such an array is called the
+*positional representation* of the instance.
```
mikepizzo, line 17:
```suggestion
Properties of an entity or a complex value MAY be represented
```
```suggestion
Structural and navigation properties of an entity or a complex value MAY be represented
as a JSON array instead of name-value pairs within a JSON object. Such an array is called the
*positional representation* of the properties.
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+list of properties that its positional representation conveys.
+
+If an instance is represented positionally:
+
+- the number of items in the array MUST equal the number of items in the
+  positional property list, and
+- item *n* of the array MUST be the value of item *n* of the positional
+  property list, formatted as described in [section ##PositionValues].
+
+A sender MUST NOT omit an item, MUST NOT add an item, and MUST NOT
```
mikepizzo, line 60:
```suggestion
A producer MUST NOT omit an item, MUST NOT add an item, and MUST NOT
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+select-list, as defined in [#OData-Protocol#ContextURL], that applies to
+that instance.
+
+The context URL of a compact payload MUST contain a select-list, and that
+select-list MUST enumerate every property conveyed positionally, at every
+level of nesting. In particular:
+
+- the select-list MUST NOT be omitted, and MUST NOT be empty;
+- the select-list MUST NOT contain the shortcut `*`, nor the shortcut
+  `{namespace}.*` for the bound operations of a type;
```
mikepizzo, line 86: why can it not contain `{namespace}.*`, along with the properties?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+- a select-item for a structured property whose value is conveyed
```
mikepizzo, line 87: should we clarify "structured navigation or structural property"?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+  positionally MUST carry a nested select-list, rather than the empty
+  parentheses that [OData-Protocol](#ODataProtocol) permits;
+- the rule of [OData-Protocol](#ODataProtocol) whereby a select-list
+  containing only expanded navigation properties implicitly selects all
+  structural properties does not apply to a compact payload.
```
mikepizzo, line 92: rather than say what doesn't apply, can we turn this around and say what must exist?  i.e., something like "For Compact JSON payloads, the select-list MUST always include all selected structural as well as expanded navigation properties."

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+select-list places the information with the party that reliably has it.
+
+Let *T* be the type of the instance and *S* the sequence of select-list
+items, in the order in which they appear in the context URL, that applies
+to the instance. The positional property list is determined as follows:
+
+1. Each item of *S* that begins with a type-cast segment --- a qualified
+   type name followed by a forward slash --- is removed from *S* unless *T*
+   is that type or is derived from it. From each such item that remains,
+   the leading type-cast segment is removed.
```
mikepizzo, line 118: I'm not sure that we need to say that the leading type-cast segment is removed. This is a logical algorithm; what difference does it make if the type cast segment is conceptually there or not?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+   ##GroupingofSelectItems]. Each group occupies exactly one position,
+   at the position of the first of its items.
+
+3. The positional property list is the resulting sequence of groups, in
+   order.
+
+Step 1 is what makes the positional property list depend on the instance
+and not only on the context URL: two instances of different types in one
+collection have different positional property lists. See [section
+##DerivedTypes].
```
mikepizzo, line 130: do we need this conceptual algorithm, or can we just say the following:

```suggestion
The positional property list for any instance is the ordered list of all properties
in the context URL that apply to the type of the instance. Thus, two instances of
different types in one collection have different positional property lists. See [section
##DerivedTypes].
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+- for an item of the form `P(nested)` or `P+(nested)`, the items of
+  `nested`;
+- for an item of the form `P`, nothing.
+
+The nested select-list MUST NOT be empty, for the reason given in [section
+##DeterminingthePositionalPropertyList]: there is no well-defined default
+to fall back on.
+
+The nested select-list determines the positional property list of the
+instances of that property, applying this section recursively.
```
mikepizzo, line 230: so, this is really saying that all nested structural and navigation properties belonging to the same structured object are grouped together and positioned at the first occurrence of such a property, right?  I got a little lost in the algorithm.

I dislike being quite so tied to the exact syntax of the contextUrl, and would rather describe the intent than an algorithm for string parsing.

```suggestion
A select-list may contain multiple items that address different structural and
expanded properties of the same complex-typed property. The positional
representation groups all nested structural and navigation properties
belonging to the same complex object together as a nested select-list,
positioned at the first occurrence of a such a property in the select-list.
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+}
+```
+The positional property list of each `Customer` is (`Name`, `Address`),
+and the positional property list of each `Address` is (`City`,
+`PostalCode`).
+:::
+
+::: example
+Example ##ex: a bare item and a suffixed item addressing the same
+navigation property form one group and therefore one position
```
mikepizzo, line 255: I don't understand the "bare" and "suffixed" items referred to in this example.

mikepizzo: Okay; I get it -- we're talking about DirectReports and recursive DirectReports with an explicit select list?  In this case, doesn't the first DirectReports just refer to including the navigation link? Should we include that the DirectReports have a nested select list with FirstName and LastName? Right now the example makes it appear that those nested properties are lost.

mikepizzo (changes requested):
a lot of mostly minor comments

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+cannot be placed in the positional property list at all. The containing
+instance is no help, since it may itself be an array.
+
+This format introduces a single construct for all of them, used uniformly
+wherever a value would otherwise appear.
+
+A *wrapper object* is a JSON object whose name/value pairs are
+
+- annotations and control information that apply to a value,
+- optionally, that value itself, under the reserved name `$`, and
```
mikepizzo, line 278:
```suggestion
- optionally, that positional values of the instance, under the reserved name `$`, and
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+- any other JSON object is the representation defined by
+  [OData-JSON](#ODataJSON).
+
+The value under `$` is whatever representation applies to that value: the
+positional representation of a structured instance, a JSON object as
+defined in [OData-JSON](#ODataJSON), a JSON array for a collection, or a
+JSON primitive where the wrapper stands at the position of a primitive
+property. It is not restricted to arrays --- a wrapper carrying an
+annotation on a primitive property holds that primitive under `$`.
+
```
mikepizzo, line 301:
```suggestion
An object with a `$` property is a wrapper object in which all of the
positional properties are represented by the property named `$` and any
other properties, control information, or  annotations are represented as
defined by [OData-JSON](#ODataJSON).
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+  "$": [
+    ["Alfreds Futterkiste", ["Berlin", "12209"]],
+    ["Ana Trujillo", ["México D.F.", "05021"]]
+  ]
+}
+```
+The positional property list of each `Customer` is (`Name`, `Address`),
+and the positional property list of each `Address` is (`City`,
+`PostalCode`).
+:::
```
mikepizzo, line 251: should we add a property between Address/City and Address/PostalCode to show the rearrangement, i.e.:


```suggestion
GET ~/Customers?$select=ID,Address/City,Name,Address/PostalCode
```
```json
{
  "@context": "$metadata#Customers(ID,Address/City,Name,Address/PostalCode)",
  "$": [
    ["ALFKI", ["Berlin", "12209"],"Alfreds Futterkiste"],
    ["ANATR", ["México D.F.", "05021"],Ana Trujillo"]
  ]
}
```
The positional property list of each `Customer` is (`ID, `Address`,`Name`),
and the positional property list of each `Address` is (`City`,
`PostalCode`).
:::
```

File: odata-compact-json-format/4 Compact Representations.md [hunk 45]
```diff
@@ -0,0 +1,443 @@
...
+navigation property form one group and therefore one position
+```
+$metadata#Employees/Sales.Manager(DirectReports,DirectReports+(FirstName,LastName))
+```
+The positional property list of each `Manager` is (`DirectReports`).
+:::
+
+## ##subsec The Wrapper Object
+
+A positional representation is a JSON array, and a JSON array has no
```
mikepizzo, line 264:
```suggestion
A positional property representation is a JSON array, and a JSON array has no
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+name/value pairs. Anything that has to be conveyed *by name* alongside a
+value therefore has nowhere to go: the annotations and control information
+that [OData-JSON](#ODataJSON) attaches to the object representing an
+instance, or to the object containing a property; and any property that
+cannot be placed in the positional property list at all. The containing
+instance is no help, since it may itself be an array.
+
+This format introduces a single construct for all of them, used uniformly
+wherever a value would otherwise appear.
```
mikepizzo, line 273: I'm not sure this is needed. We don't normally include this must justification/explanation -- we just explain the intended design/behavior.  In this case, I think it's pretty straightforward why we chose what we chose.


```suggestion
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+The value under `$` is whatever representation applies to that value: the
+positional representation of a structured instance, a JSON object as
+defined in [OData-JSON](#ODataJSON), a JSON array for a collection, or a
+JSON primitive where the wrapper stands at the position of a primitive
+property. It is not restricted to arrays --- a wrapper carrying an
+annotation on a primitive property holds that primitive under `$`.
+
+Properties carried by name in a wrapper object MUST NOT include a property
+that occupies a position in the instance's positional property list, and
+MUST follow the value, as required by [section
```
mikepizzo, line 304: why must the property carried by name follow the value?  what value (the $ property?)

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+  [OData-JSON](#ODataJSON).
+
+The value under `$` is whatever representation applies to that value: the
+positional representation of a structured instance, a JSON object as
+defined in [OData-JSON](#ODataJSON), a JSON array for a collection, or a
+JSON primitive where the wrapper stands at the position of a primitive
+property. It is not restricted to arrays --- a wrapper carrying an
+annotation on a primitive property holds that primitive under `$`.
+
+Properties carried by name in a wrapper object MUST NOT include a property
```
mikepizzo, line 302: would it be clearer to say "identified by name" rather than "carried by name"?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+that occupies a position in the instance's positional property list, and
+MUST follow the value, as required by [section
+##PayloadOrderingConstraints]. Carrying properties by name is what allows
+an instance of an open type to keep its positional representation while
+conveying dynamic properties that could not be placed in the select-list;
+see [section ##OpenTypesandDynamicProperties].
+
+A wrapper object MAY appear wherever a value may appear:
```
mikepizzo, line 310: where-ever a JSON value may appear, or a positional property value may appear, or?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+- as the message body, in which case it carries the payload's
+  [`context`](#ControlInformationcontext) and the payload's content;
+- at a position in a positional representation, in which case it carries
+  the annotations of the property at that position and the property's
+  value;
+- as an item of a collection, in which case it carries the annotations of
+  that member of the collection and the member itself.
```
mikepizzo, line 318: not just annotations, but also control information and dynamic properties, right?


```suggestion
- at a position in a positional property representation, in which case it represents the
  the annotations, control information, positional and dynamic properties
  of the property at that position
- as an item of a collection, in which case it represents the annotations, control
information, positional and dynamic properties of that member of the collection.
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+A wrapper object that carries no value denotes a property or instance that
+has no value, as distinct from one whose value is null. The empty JSON
+object `{}` carries neither annotations nor a value and therefore denotes
+"no value"; it is used at the position of a selected dynamic property that
+an instance does not have, see [section ##OpenTypesandDynamicProperties].
+Note that `{}` satisfies the second bullet above vacuously rather than by
+construction: it is treated as a wrapper carrying no value by convention,
+being the only reading that is useful.
```
mikepizzo, line 327:
```suggestion
The empty JSON object `{}` is treated as a wrapper object with no value;
it is used at the position of a selected dynamic property that
has no value for the instance. See [section ##OpenTypesandDynamicProperties].
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+The name of the value in a wrapper object is `$`, in every position in
```
mikepizzo, line 329:
```suggestion
The name of the positional property values in a wrapper object is `$`, in every position in
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+:::
+
+Note the consequence of the two preceding rules: for a
+collection-valued structured property, the value at the position is an
+array of arrays. A receiver that knows the positional property list also
+knows, from the metadata, whether a property is collection-valued, and can
+therefore distinguish the two nestings without ambiguity.
+
+A property that has no value at all -- as opposed to a property whose value
+is null -- occurs when only annotations were requested for it, for example
```
mikepizzo, line 440:
```suggestion
is null -- occurs when only annotations or control information were requested for it, for example
```

odata-compact-json-format/4 Compact Representations.md
mikepizzo: savings is in the array response, so would likely use value for singleton and $ for array.

odata-compact-json-format/2 Compact JSON Format Design.md
mikepizzo: shouldn't use "positional properties" here -- should say something like "known properties" that anyone familiar with the json specification understands.

odata-compact-json-format/3 Requesting the Compact JSON Format.md
HeikoTheissen: > doing so adds this as a requirement to existing services

Perhaps Hubert was referring to services that do advertise support for compact JSON in general but not support a particular compact JSON payload?

But why not just let the service respond like a compact-JSON-unaware service would: with 400 Bad Request?

odata-compact-json-format/3 Requesting the Compact JSON Format.md
HeikoTheissen: Doesn't RFC2045 (quoted by Mike below) apply here, which states:
> MIME implementations must ignore any parameters whose names they do not recognize.

Wouldn't the service simply ignore the `compact=true`? To prevent clients from falling into this trap, we should specify
> Clients are advised that services unaware of compact JSON will treat `Accept: application/json;compact=true` like `Accept: application/json` (because RFC2045 instructs them to ignore the unknown `compact=true`). Therefore, a client SHOULD NOT send requests with `Accept: application/json;compact=true` unless it has established … [continues like below]

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+The *first segment* of a select-item is the item with any `(...)` or
+`+(...)` suffix removed, truncated before the first forward slash (`/`).
+Leading type-cast segments have already been removed by step 3 of [section
+##DeterminingthePositionalPropertyList] and so do not occur here.
+
+Two items of *S* belong to the same group if and only if their first
+segments are equal. The group occupies the position of the first of its
+items.
+
+If the property addressed by the first segment of a group is of a
```
HeikoTheissen, line 216:
```suggestion
If the property addressed by the first segment `P` of a group is of a
```
You use the variable `P` below, I assume this is the first segment.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+structured type, the *nested select-list* of that group is formed by
+concatenating, for each item of the group in order:
+
+- for an item of the form `P/rest`, the item `rest`;
+- for an item of the form `P(nested)` or `P+(nested)`, the items of
```
HeikoTheissen: Can `nested` be empty (in which case the item `P()` remains)?

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+## ##subsec Determining the Positional Property List
+
+The positional property list of an instance is determined by the
+select-list, as defined in [#OData-Protocol#ContextURL], that applies to
+that instance.
+
+The context URL of a compact payload MUST contain a select-list, and that
+select-list MUST enumerate every property conveyed positionally, at every
+level of nesting. In particular:
```
HeikoTheissen, line 82: Can nested context URLs also define a positional property list? For example, if not every customer has addresses:
```json
{"@context": "$metadata#Customers(ID,Name)",
 "$": [
  {"$": ["ALFKI", "Alfreds Futterkiste"],
   "Addresses@context": "#Addresses(Street,City)",
   "Addresses": [
    ["Friedrichstraße", "Berlin"],
    ["Bahnhofstraße", "Walldorf"]
   ]
  }, ...
 ]}
```

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+- a JSON object containing a name/value pair named `$` is a wrapper object;
+- a JSON object all of whose name/value pairs are annotations or control
+  information is a wrapper object that carries no value;
+- any other JSON object is the representation defined by
+  [OData-JSON](#ODataJSON).
+
+The value under `$` is whatever representation applies to that value: the
+positional representation of a structured instance, a JSON object as
+defined in [OData-JSON](#ODataJSON), a JSON array for a collection, or a
```
HeikoTheissen: How easy is it to distinguish between positional representation and collection if the value under `$` is of the form `[{…},{…},…]`? The nested objects could be values of a positional property or items of a collection, given as wrapper objects.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+property `Address` share one position, which holds the positional
+representation of the complex value
+```
+GET ~/Customers?$select=Name,Address/City,Address/PostalCode
+```
+```json
+{
+  "@context": "$metadata#Customers(Name,Address/City,Address/PostalCode)",
+  "$": [
+    ["Alfreds Futterkiste", ["Berlin", "12209"]],
```
HeikoTheissen, line 243: If `Address` was collection-valued, then the context URL for
```
GET ~/Customers?$select=Name,Address($select=City,PostalCode)
```
would be
```
$metadata#Customers(Name,Address(City,PostalCode))
```
and the representation would be
```json
["Alfreds Futterkiste", [["Berlin", "12209"]]]
```
Correct?

A consumer of compact JSON payloads can parse this only with knowledge about the cardinality of the `Address` property, which it obtains from the $metadata. This reliance on $metadata should be mentioned in the text somewhere, because it wasn't the case for standard JSON format.

odata-compact-json-format/4 Compact Representations.md
HeikoTheissen: "Dynamic properties of a property" is not needed. A wrapper object in position N of a positional property representation carries information about the N-th property:
- context (not: control) information for that property, for example, its `@type`
- annotations of that property
- the value of that property (this can be a structured value with dynamic properties, but then we are one nesting level deeper).

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+positionally at the root of the message body therefore uses `$`.
+
+Producers of compact payloads SHOULD use `$` wherever this document
+permits a choice.
+
+This restriction is what keeps the two representations distinguishable.
+Were `value` also the wrapper's value name at the root of a message body
+representing a single entity, then
+`{"@context": "…#Customers/$entity", "value": […]}` would be at once the
+positional representation of an entity and the
```
HeikoTheissen, line 355: This would _not_ be a valid positional representation, because the context URL lacks the select-list that defines the positional property list.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+has no room for one. Everything else in this document is expressed in
+terms of these two.
+
+## ##subsec Positional Representation
+
+A structured instance -- an entity or a complex value -- MAY be represented
+as a JSON array instead of a JSON object. Such an array is called the
+*positional representation* of the instance.
+
+The items of the array are the values of the instance's properties. A
```
HeikoTheissen, line 19: But the items of an array can also be items of a collection. In order to distinguish these two, parsers must know the cardinalities of properties from the $metadata, see https://github.com/oasis-tcs/odata-specs/pull/2270#discussion_r3979658023

Insert another example after example 4 that illustrates this.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+as described in [section ##PositionValues].
+
+Note that this is a stricter requirement than the one
+[OData-JSON](#ODataJSON) places on a JSON object representation, where a
+sender may omit a property whose value it does not wish to transmit. In a
+positional representation there is no way to omit a value without
+shifting every subsequent value, so the positional property list must be
+transmitted in full. If a sender wishes to transmit fewer properties, it
+narrows the positional property list -- by narrowing the select-list in the
+context URL -- rather than shortening the array.
```
HeikoTheissen, line 72:
```suggestion
context URL.
```
Because the array is also shortened if the select-list is narrowed.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+  the annotations of the property at that position and the property's
+  value;
+- as an item of a collection, in which case it carries the annotations of
+  that member of the collection and the member itself.
+
+A wrapper object that carries no value denotes a property or instance that
+has no value, as distinct from one whose value is null. The empty JSON
+object `{}` carries neither annotations nor a value and therefore denotes
+"no value"; it is used at the position of a selected dynamic property that
+an instance does not have, see [section ##OpenTypesandDynamicProperties].
```
HeikoTheissen: Are the following equivalent for select-list `(ID,Name)`?
```json
{"$": ["ALFKI", {}],
 "Name@Core.Permission": "None"}
```
```json
["ALFKI", {"@Core.Permission": "None"}]
```
Or does the first violate the following rule?
> Properties carried by name in a wrapper object MUST NOT include a property that occupies a position in the instance’s positional property list

Does this merit an example?

odata-compact-json-format/4 Compact Representations.md
HeikoTheissen: It must follow the `$` member in the wrapper object.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,443 @@
...
+
+The value under `$` is whatever representation applies to that value: the
+positional representation of a structured instance, a JSON object as
+defined in [OData-JSON](#ODataJSON), a JSON array for a collection, or a
+JSON primitive where the wrapper stands at the position of a primitive
+property. It is not restricted to arrays --- a wrapper carrying an
+annotation on a primitive property holds that primitive under `$`.
+
+Properties carried by name in a wrapper object MUST NOT include a property
+that occupies a position in the instance's positional property list, and
```
HeikoTheissen, line 303: Does this only apply to the property itself, or also to its annotations/control information?

In other words: Is the following fobidden (assuming select-list `(ID,Name)`)?
```json
{"$": ["ALFKI", "Alfred"],
 "Name@Core.Permission": "Read"}
```
Or is it equivalent to
```json
["ALFKI",
 {"$": "Alfred", "@Core.Permission": "Read"}]
```

odata-compact-json-format/5 Common Characteristics.md
```diff
@@ -0,0 +1,99 @@
...
+              IEEE754Compatible=true;streaming=true
+```
+:::
+
+## ##subsec Message Body
+
+Each message body is represented as a single JSON object, as defined in
+[OData-JSON](#ODataJSON), with the single exception described in [section
+##MessageBodyofaRequest]. A response message body is always a JSON
+object: the positional representation applies to the instances *within*
```
mikepizzo, line 30:
```suggestion
object: the positional representation applies to the property instances *within*
```

odata-compact-json-format/5 Common Characteristics.md
```diff
@@ -0,0 +1,99 @@
...
+
+- the representation of an [entity](#Entity), an [entity
+  reference](#EntityReferences) or a [complex value](#ComplexValue) that
+  is not represented positionally, or
+- a [wrapper object](#wrapperobject), whose value is the correct
+  representation for the payload's content.
+
+The name of the value in a wrapper object is `$`. The name `value` is also
+recognized, but only in those message bodies in which
+[OData-JSON](#ODataJSON) itself uses it; see [section ##TheWrapperObject].
```
mikepizzo, line 43: Alternatively, could we say that the object is *always* a representation of an entity, ref, or complex type, and that the positional properties may either be represented as individual name/value pairs or as a single positional  wrapper obect named `$`?



```suggestion
This properties of an entity, entity reference, or complex type are represented as either

- individual name/value pairs
- 

```
```suggestion
An entity, entity reference, or complex type is represented as either

- the standard OData JSON representation of an [entity](#Entity), an [entity
  reference](#EntityReferences) or a [complex value](#ComplexValue) with 
  properties represented as individual name/value pairs
- the standard OData JSON representation of an [entity](#Entity), an [entity
  reference](#EntityReferences) or a [complex value](#ComplexValue) with 
  all positional properties represented a single array-valued property whose
  name is `$`
- an array of positional property values
```

odata-compact-json-format/5 Common Characteristics.md
```diff
@@ -0,0 +1,99 @@
...
+
+::: example
+Example ##ex: a message body containing a collection of entities
+represented positionally
```
mikepizzo:
```suggestion
whose properties are represented positionally
```

mikepizzo, line 47:
```suggestion
represented as arrays of positional property values.
```

odata-compact-json-format/3 Requesting the Compact JSON Format.md
HeikoTheissen: We agree this is not a trap.

odata-compact-json-format/4 Compact Representations.md
HeikoTheissen: It is forbidden.

odata-compact-json-format/4 Compact Representations.md
HeikoTheissen: For the purpose of computing the positional property list, select-lists like (Address/City, Address/PostalCode) are converted to nested.

odata-compact-json-format/4 Compact Representations.md
```diff
@@ -0,0 +1,691 @@
...
+Where the value of a property is a wrapper object, all of that property's
+annotations and control information MUST be carried in it, and none MUST
+appear in a name/value pair prefixed with the property name. This is the
+rule of [section ##TheWrapperObject] --- everything that applies to one
+property is in one place --- applied to the object representation. The two
+forms convey the same thing, and a consumer MUST read them as equivalent.
+
+::: example
+Example ##ex_grouped: the annotated property of [example
+##ex_propannotation], in an instance that is not represented positionally:
```
HeikoTheissen:
```suggestion
##propannotation], in an instance that is not represented positionally:
```

odata-compact-json-format/8 Request Payloads.md
```diff
@@ -0,0 +1,223 @@
...
+be updated. A property that is not in the positional property list is
+absent from the request in the sense of
+[OData-Protocol](#ODataProtocol) and is left unchanged. A property that
+is in the positional property list with the value `null` is set to null.
+
+The `context` control information is therefore what makes `PATCH`
+expressible in this format, and is one reason [section
+##DeterminingthePositionalPropertyListinRequests] requires it of every
+compact request body: without it there would be no way to say which
+properties a `PATCH` leaves alone.
```
HeikoTheissen: This is not specific to Compact JSON, suggest to delete this paragraph and the example.

PR #2271, opened 2026-08-20
That would be too easy

HeikoTheissen:
This PR extends the required "build (24.x)" check so that it commits and pushes changes to the generated files in the `docs` folder instead of failing if there are any such changes.

The drawback is that the extra commit does not trigger another execution of the workflow so that the required check is missing. The commit message will contain the statement
> Go to https://github.com/oasis-tcs/odata-specs/actions/workflows/nodejs.yml
> and run the workflow manually on refs/heads/«name of branch»
> in order to have the required build (24.x) check executed.

Is this helpful?

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,6 +1,6 @@
 name: Node CI
 
-on: push
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,6 +1,6 @@
 name: Node CI
 
-on: [push, pull_request]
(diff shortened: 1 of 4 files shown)


Hubert-Heijkers (commented):
Heiko, help me understand, if we enforce build after every push why would we still need that build check? Perhaps I don't know what that build check does, I got the message because I hadn't build as part of what I pushed, but this change would implicitly fix that for me not?

HeikoTheissen:
If we don't enforce the check, then the build might fail (because of a syntax error in the Node.js program, say), and the PR would still be mergeable. The purpose of the check is not only to ensure a build has been carried out, but also that the build program itself works.

PR #2285, opened 2026-09-08
Do not treat groupby key fields specially

HeikoTheissen:
Instances in an output set shall retain their metadata (such as etags) and navigation properties only if they result from a transformation like `identity` or `groupby((«navigation property»))`, not from grouping by their key fields.

**Rationale:** Detecting whether the properties in a `groupby` contain all key fields is unnecessary extra work. The desired transformation effect can more naturally be expressed through  `identity` or `groupby((«navigation property»))`.


ralfhandl (dismissed):

SebastianDobrzynski:
Would `groupby((<<navigation property>>))` then have a different  behaviour to selecting every structural property of the navigation property  individually in the groupby or would this be equivalent? I mean `groupby((NavProp/Prop1,NavProp/Prop2,NavProp/Prop3...))`  until every structural property of NavProp is selected.
Honestly, I feel like it would feel kind of weird if both requests select the same structural properties, but differ in whether they have the metadata attached or not. We would also currently treat both requests as the same.

HeikoTheissen:
> Would `groupby((<<navigation property>>))` then have a different behaviour to selecting every structural property of the navigation property individually in the groupby or would this be equivalent? I mean `groupby((NavProp/Prop1,NavProp/Prop2,NavProp/Prop3...))` until every structural property of NavProp is selected. Honestly, I feel like it would feel kind of weird if both requests select the same structural properties, but differ in whether they have the metadata attached or not. We would also currently treat both requests as the same.

Yes, that would differ. Grouping by every structural property of the related entity would not preserve the ETags, or navigation properties of the related entity. (Detecting every structural property in `groupby` is as cumbersome as detecting every key property.)

PR #2289, opened 2026-09-15
Repeatable Requests CSD02

HeikoTheissen:
Translated the CS01 Word document to Markdown as a starting point for CSD02.


SebastianDobrzynski (approved):
So far just the 1.0 version as .md as a starting point for CSD02.  Looks fine for now.

